PR #55, opened 2019-08-13
Added basic routes, controller, services for authentication (register/login)

nclairesays:


nclairesays:
currently unable to connect to the ms sql server but i set up routing structure for auths with brypt. Planning to use JWT tokens instead of sessions.

Issue #26, opened 2019-07-31
Create a CRUD feature for an admin to enter and maintain a list of frequently asked questions.


entrotech:
### Overview 
Provide a page for a user with the admin role to view a list of FAQs, create new ones and edit existing ones.
### Action items
- [x] Creating the crud feature
- [ ] Design the React Components for listing existing FAQs
- [ ] Design React Form(s) for Adding and Editing an FAQ and connect with the list.
- [ ] Add appropriate menu items which are enables only for an admin user.

kkchu791:
I'm interested to take this one on.

nicholaspung:
Sent a PR. Feature done!

nicholaspung:
Front End
- Display FAQs with update/delete functionality
- Able to add FAQs

ExperimentsInHonesty:
Needs to be integrated.  new issue will be created

KPHowley:
closed as we have duplicate issues for FAQ screen  #76 and #42 and #604

PR #78, opened 2019-09-16
Added first draft of Git instructions and FAQ front-end

nicholaspung:
FAQ Front-End is displayed on "Admin" page where a list of FAQs are rendered.
Depending if a user is an admin, add/update/delete functionality will display.

Code follows style guidelines.
All manual tests all pass.
Comments left in place are for future connections (i.e. checking if user is an admin)


nicholaspung:
This PR is for #26

Issue #123, opened 2019-12-24
Add User Notification capabiltiy to TDM like "Toasts"


entrotech:
When the application needs to notify a user that something has happened without a modal dialog, we need some sort of transient notification (i.e. a pop-up panel that displays a message for a few seconds then vanishes). This needs to be implemented across the application, and preferably implemented in an application-standard way, and available across the appllication (maybe via React Context?)  This is a pre-requisite to a good, complete implementation of the registration/login and forgot password features.

Issue #71, opened 2019-09-03
Phase 2: Develop login front-end component


nclairesays:
Backend login routes have been created via issue #48 . 

Login form to be created. send email/password to backend which will receive jwt token and be stored into local storage.
-------------
This will be a fairly typical SPA user registration and login.  Adding roles, etc. will be a separate issue.

- [x] Back-End Web API implementation
- [x] Front-End functionality
- [x] Front-End User Notifications to replace console.log
- [x] Cosmetic clean-up

entrotech:
I expect to implement Login/Logout and Forgot Password features similar to what we have in Food Oasis and demonstrated in this video:

https://www.loom.com/share/3de25f3c92e84d2b927a1e87cc76f6a7

entrotech:
Also added a GitHub markdown page describing in detail how the registration/login features are intended to work: https://github.com/hackforla/tdm-calculator/blob/master/register.md

entrotech:
I believe this is working, though it needs to be thoroughly tested per the register.md process description, including all the "unhappy paths". In the Login.js component and ConfirmEmail.js component, I put a TODO comment where we need to add user notifications (see Issue #123), to let the user know what happened and/or what to do next. I'll let somebody else pick up this issue and complete it :)

Of course, the appearance and styling also needs work!

ExperimentsInHonesty:
@nclairesays currently the live registration page doesn't look like the figma registration page

![Screen Shot 2020-01-02 at 2 09 36 PM](https://user-images.githubusercontent.com/37763229/71696325-8edebf00-2d69-11ea-9236-f8d05be7261a.png)

![Screen Shot 2020-01-02 at 2 08 39 PM](https://user-images.githubusercontent.com/37763229/71696268-6bb40f80-2d69-11ea-8c0f-73fd732fb760.png)

Can you change the live version to match?

ExperimentsInHonesty:
@nclairesays currently the live login  page doesn't look like the figma registration login page

![Screen Shot 2020-01-02 at 2 13 48 PM](https://user-images.githubusercontent.com/37763229/71696574-365bf180-2d6a-11ea-8a56-4a1f2ffd7838.png)

![Screen Shot 2020-01-02 at 2 16 07 PM](https://user-images.githubusercontent.com/37763229/71696690-79b66000-2d6a-11ea-9b1c-4e140eca3d64.png)

Can you change the live version to match Nicole's Figma?

nclairesays:
[...]

Issue #44, opened 2019-08-06
Phase 2: Develop user registration functionality


entrotech:
This will be a farily typical SPA user registration and login.  Adding roles, etc. will be a separate issue.

- [x] Back-End Web API implementation
- [x] Front-End functionality
- [x] Front-End User Notifications to replace console.log
- [ ] Cosmetic clean-up

nclairesays:
I've created the backend registration routes.
I"m going to go ahead and create the front end registration component.

@pelmini is working on the Login Frontend component

entrotech:
Here's a video showing how registration works in Food Oasis - this is pretty much what I expect to implement for TDM.

https://www.loom.com/share/e11a92f6878d48adac9ed65c947770af

entrotech:
Also added a GitHub markdown page describing in detail how the registration/login features are intended to work: https://github.com/hackforla/tdm-calculator/blob/master/register.md

entrotech:
I believe this is working, though it needs to be thoroughly tested per the register.md process description, including all the "unhappy paths".  In the Register.js component, I put a TODO comment where we need to add user notifications, to let the user know what happened and/or what to do next. I'll let somebody else pick up this issue and complete it :)

nclairesays:
This is already moved from review to done. Closing this issue.

PR #131, opened 2019-12-29
Replaced console logs with toasts for Registration, Login, ConfirmEmail

nclairesays:
Utilized the useToast context to add toast messages to Registration, Login, and ConfirmEmail components.  Minor updates to toast css.

Relevant to issues: 
#123 (utilized useToast context), 
#44 (Front-End User Notifications to replace console.log)
#71 (Front-End User Notifications to replace console.log)

### DIFF
--- a/client/src/components/ConfirmEmail.js
+++ b/client/src/components/ConfirmEmail.js
@@ -1,6 +1,7 @@
 import React, { useState, useEffect } from "react";
 import { Redirect, withRouter } from "react-router-dom";
 import * as accountService from "../services/account-service";
+import { useToast } from "../contexts/Toast";
 
 const ConfirmEmail = props => {
   const { history } = props;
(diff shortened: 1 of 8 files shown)


client/src/components/ConfirmEmail.js
```diff
@@ -40,6 +40,7 @@ const ConfirmEmail = props => {
       ) : confirmResult.success ? (
         <Redirect to={`/login/${email}`} />
       ) : emailSent ? (
+        // TODO: CHECK ON THIS CODE - How do we test this to see this?? -- Claire
```
entrotech: You would test this by setting up an invalid or expired token, either by waiting an hour after registration for the token to expire, or by manually deleting the token from the database before clicking on the link in the email.  Then, when you get to this page, the result should be an expired or invalid token, an the user should be instructed to push the button to generate a fresh token and send another confirmation email, which should cause this code to execute.

nclairesays:
@entrotech how about 10 seconds? I just thought the original 5 seconds was just way too short especially if you're trying to read the longer error/warning messages and trying to figure out what you're supposed to do. It was also because the original color scheme made it difficult for me to even find/see the toast and it disappeared too quickly by the time I got to reading it. Personally, I would not want the error/instructional messages to disappear at all for login/registration because sometimes I'd want to refer back to the message.  And i think many sites have it that way where it doesn't disappear.

An idea for if we end up using toasts more in the other locations: since the toast duration is currently set up to affect all toasts globally, maybe in the future we can add an optional prop to customize duration?

nclairesays:
@entrotech also, i think i understand your comment earlier about using toast inside the Effect for about and contact. I think that was more so for @nicholaspung to test out the toasts. We can probably delete the toasts from those two components since yes, we may not need any other real examples where we'd need a toast inside an useEffect.

PR #154, opened 2020-01-06
Revert "Push to /login instead of /admin if login successful"

ExperimentsInHonesty:
Reverts hackforla/tdm-calculator#153


nclairesays:
was supposed to be /calculation not /login
here's the correct change: https://github.com/hackforla/tdm-calculator/commit/5e5e2aec7a5691377428e9f6b35d32165930449c

PR #169, opened 2020-01-08
Request password reset works. Fixed login bug.

bfojas:
Request password reset works. Fixed login bug.
Screenshot from 2020-01-07 21-00-11


nclairesays:
#71 password recovery

PR #186, opened 2020-01-12
Fixed conditional render on password reset request. Fixed firefox render of login buttons.

bfojas:
Screenshot from 2020-01-11 21-50-39
Screenshot from 2020-01-11 21-51-09

- Adjusted CSS on login component to include height for buttons. 
- Password reset request was working, but not sending the proper response to the front end. Conditional render to alert the user to check email was not being displayed because the server response didn't match the conditions to change the render.


bfojas:
@entrotech Hold off on this PR for a bit. I just realized why the'forgot password' API endpoint was sending the sendgrid service response on success (which was what I fixed). It sends that in case there is an error from the sendgrid request. So now this PR would have a bad response if the sendgrid request has an error. I'm going to make some quick changes and push again.

bfojas:
OK, done. That fix was quicker than I thought it would take me.This PR should be ready.

bfojas:
> @entrotech Hold off on this PR for a bit. I just realized why the'forgot password' API endpoint was sending the sendgrid service response on success (which was what I fixed). It sends that in case there is an error from the sendgrid request. So now this PR would have a bad response if the sendgrid request has an error. I'm going to make some quick changes and push again.

OK, done. That fix was quicker than I thought it would take me.This PR should be ready.

Issue #168, opened 2020-01-08
Adjust Transporation Demand Measures page to highlight measures, and display possible & earned points more clearly 


KPHowley:
### Overview
Adjust Transportation Demand Measures page to highlight each measure as they are selected and to improve the visibility of the two columns with the possible points for each measure and the earned points. 

### Action Items

- [x]  add a column header to the possible points column that says "possible points"
- [x]  Add "pts" next to each number in the possible points column
- [x]  add a column header to the earned points column that says "earned points"
- [x]  make all values in the earned points column invisible until a measure is selected. 
- [x]  Highlight each measure and the possible points with a line or other demarcation when it is hovered over.

nclairesays:
Here's what I came up with for now because I did not have a design to go off of. It will be difficult to put a single header above the strategies without major architectural changes as it will affect other "pages" of the wizard. 
![image](https://user-images.githubusercontent.com/45305029/72236711-d3016900-358c-11ea-8aaf-810207a3c14e.png)

nclairesays:
@entrotech  Due to the fact that this component/page affects the previous page too, can you add "units" in the database for "parking required" and "project level"  Otherwise, it returns as null. For now, I added a conditional that says if the units are null/undefined then the units will be an empty string, but it kinda makes the page a little awkward. see image below.
![image](https://user-images.githubusercontent.com/45305029/72236791-178d0480-358d-11ea-853c-69934e923129.png)

entrotech:
@nclairesays  Your screen shot is for Target Point calculation page. This issue is for the Transportation Demand Strategies page.  The Target Point Calculation page is going to be different than the rest, and Nicole has an issue to come up with a wireframe for what it should look like (Issue #142) before we make any code changes.  Even so, I don't really understand what you are trying to accomplish or what you are asking for.

bfojas:
@nclairesays @entrotech 
 Claire, are you asking about the having filler '0's for the empty earned values? I made a change to fill with '0' rather than the empty string (on my local just now), to see what it looks like. Is this what you are trying to get?

![Screenshot from 2020-01-13 17-00-39](https://user-images.githubusercontent.com/43628838/72304748-51f6b000-3626-11ea-8aa6-c30696c2fa35.png)

ExperimentsInHonesty:
@fyliu - Please annotate this issue with the notes you had for @nclairesays about this.

Issue #142, opened 2020-01-04
UI/UIX Wireframe: Target Point Calculation Page 


entrotech:
### Overview
Design the target point calculation page in the Calculation wizard.

### Action Items
- Design in Figma
- Assign someone to review
- Create New Issue for front end development 

### Resources/Instructions
Figma Wireframe Link:
https://www.figma.com/file/qNgPc5lttpf0SWSwkbbWyT/TDM-Calculator?node-id=0%3A1
Follows up issue #18

KPHowley:
<img width="1440" alt="A7 Target point Calculation@2x" src="https://user-images.githubusercontent.com/49739276/72689779-24987f00-3aca-11ea-9516-bd092f395422.png">
Here's a proposal for a slightly updated design.
@nekobox - Can you review, and any changes. If you can get back to me by tomorrow 1/20 I'd appreciate it!

KPHowley:
@ExperimentsInHonesty - if you want to review and share, I can then close this out and make a new issue for front end development

PR #188, opened 2020-01-13
highlight measures, display points, headers, style refactor

nclairesays:
#168 Here's what I came up with for now. It will be difficult to put a single header above the strategies without major architectural changes as it will affect other "pages" of the wizard.  

image


nclairesays:
@mmorgs

fyliu:
Tagging #142 and #200 for continuation work.

Issue #201, opened 2020-01-16
New Project link doesn't work within a project


fyliu:
### Overview
New Project link doesn't create a new project when there's already one open.

Currently, with the deep-linking (#141 ) of wizard pages, clicking the link goes to the 1st page. It wasn't working before that either. It would just not do anything.

### Action Items
- [ ] Create a new project when New Project link is clicked within an open project, and go to page 1.

ExperimentsInHonesty:
Decision by Bonnie, Bradley and Kevin to 
conditionally render the new project link, based on if they are on a project.
if on project - don't show the new project link
if not on project - show the new project link

nclairesays:
@ExperimentsInHonesty @bfojas @KPHowley  what is the flow then If they are on a project but want to scrap their project and create a new project?

KPHowley:
Currently the way would be to leave the wizard  by going to a page such as "projects" and then going back. Ultimately we would most likely want a "start over" or some sort of abandon and create a new project button

I think some confusion is caused by the entire wizard view being called "new project" in the menu, and this should be a topic for further  discussion.

nclairesays:
Ok, for now, i'll conditionally render the new project link for now

PR #204, opened 2020-01-21
Conditionally render New Project Link in Nav

nclairesays:
#201 
Added useState to App.js to set/get a boolean variable (isCreatingNewProject, setIsCreatingNewProject)
pass down getter variable to Header > Navbar
pass down setter variable to TDM Calculation Container

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -26,24 +26,10 @@ const useStyles = createUseStyles({
   }
 });
 
-const setTokenInHeaders = () => {
-  axios.interceptors.request.use(
-    config => {
-      let token = localStorage.getItem("token");
-      if (token) {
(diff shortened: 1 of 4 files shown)


nclairesays:
image

image

client/src/App.js
```diff
@@ -68,16 +54,32 @@ const App = props => {
...
+          config.headers["Authorization"] = `Bearer ${token}`;
+        }
+        return config;
+      },
+      error => Promise.reject(error)
+    );
+  };
+
+  setTokenInHeaders();
+
```
fyliu: This looks like something unrelated to the ticket. Can you move it back to above the App declaration where it used to be?
It looks like something used for some manual testing purpose.

Issue #116, opened 2019-12-20
Implement Client-Side Validation for Rules with numeric inputs or measures


entrotech:
Some input rules have a range of valid numeric inputs. Implement client-side validation driven by the minValue and maxValue columns in the calculationrule table.

entrotech:
Closed by PR #159

fyliu:
The min and max values is being done with html, where the up/down arrows can't get out of range. But the user is able to type in out of range values.

ExperimentsInHonesty:
@fyliu please update this issue with:
1. Progress
2. Blocks
3. Availability
4. Eta to completion

fyliu:
Progress: Code written. Testing/refactoring before merge later today.

fyliu:
Kevin found another issue where the user can enter alpha characters. Will need to limit inputs to 0-9 and "-" sign.

fyliu:
Changed the fields from number to text, which is the solution from this issue in React: https://github.com/facebook/react/issues/16554 . It's just easier to use text with manual validation.

The problem with the number type input is that "-" sign and "e" are allowed in the input box and there's no event trigger for something like "e1" or "1ee1". "1e1" would be 1 * 10^1 which does trigger an onChange event.

Suggested test inputs:
- 12
- 012 => 12
- -5 => 5 (minus sign won't show up)
- 123 => 100 (will snap to max of 100 when 100 is the max)
- e (won't show up)
- 1e1 => 11 (e will be skipped)
![2020-01-21 20 35 06 localhost 094764901c5b](https://user-images.githubusercontent.com/1160105/72865818-0e3b2080-3c8e-11ea-9951-a790a8e88acc.png)
![2020-01-21 20 37 03 localhost f6cddf8daa38](https://user-images.githubusercontent.com/1160105/72865819-0e3b2080-3c8e-11ea-898b-698ddc6cc8de.png)
![2020-01-21 20 37 55 localhost 5c8a198fe86f](https://user-images.githubusercontent.com/1160105/72865821-0ed3b700-3c8e-11ea-8f34-c92346a48c48.png)

PR #205, opened 2020-01-22
Fix number range validation

fyliu:
Fix min/max validation of number input

Relates to issue #116


fyliu:
@ExperimentsInHonesty This code should be good. Please merge.
I have testing notes and screenshots in the issue.

Issue #174, opened 2020-01-10
User Registration: Need prompt to check email


entrotech:
When a user registers successfully, they need to be prompted with a notification to check their email. This used to work until PR #163   , which replaced the toaster-style notification with a message that shows up on the Register.js component.  The problem is that upon successful registration, the user is immediately redirected to the Login page and does not get a chance to see this important message.

ExperimentsInHonesty:
@bfojas says eta is 1/18 end of day

bfojas:
Fixed in PR #208

PR #208, opened 2020-01-24
Registration

bfojas:
Issue [#174](https://github.com/hackforla/tdm-calculator/issues/174)

After successful registration, a notice to check email is conditionally rendered.

Screenshot from 2020-01-23 16-02-15


fyliu:
@bfojas could you remove the commented out codes in Login.js and Register.js?
Also, I think the default prettier plugin we've been using does double quote for strings. Can you run Register.js through it again?

I could add these changes to the pull request if you want.

fyliu:
Fixes #174

PR #215, opened 2020-01-29
styled reset page

bfojas:
Reset page option styled to specifications of @ExperimentsInHonesty & @KPHowley .


fyliu:
@bfojas TdmCalculationWizard.js somehow switched to DOS style line endings(cr lf). I switched it back in the merge.

Issue #218, opened 2020-01-30
Fix validation issues that are keeping projects from saving.


ExperimentsInHonesty:
### Overview
There is a bug in the saving of projects.  Specifically, there are three fields, Project Name, Address & Description, that if the user does not fill out, the project will not save.

### Action Items
Back end:
- [x] Remove the requirement for description on the back end.

Front end:
- [x] Add an asterisk on the new project page, next to Project name and Address.
- [x] Create front end validation to keep the next button from working if they have not filled that out
- [x] Display the customary red warning when they move on to another field without filling one of those fields out.

ExperimentsInHonesty:
snipit from slack
Bradley Fojas  11:39 AM
I just went into the DB real quick and removed the 'not null' from the description column on the projects table. Let me know if that fixes anything, and I'll update the issue about it when I get a chance 
Bonnie  11:43
...your change to make the description field not null, definitely worked.  I was able to save a project without a description and only the address and title.

fyliu:
@ExperimentsInHonesty should the asterisks be on the left or right side of the field description? Right now, one of them says "Project Name (Required)".

ExperimentsInHonesty:
@fyliu  I think on the right is ideal, but it needs a space and should be red

![79YKW](https://user-images.githubusercontent.com/37763229/73485858-c20c7200-4358-11ea-8c33-229a5cb9331e.png)

ExperimentsInHonesty:
fang says eta is before next tuesday meetup 2/4/20

fyliu:
The main functionality is working. @KPHowley and I found a couple more things that need to be done. They're noted in the pull request.

fyliu:
I'm still working on these 2 things:
- [ ] The forward button should be set initially.
- [ ] The onChange trigger for the required input fields should update forward button.

It's working but I need to do some clean up.

fyliu:
This is still not working. Steps to reproduce
1. Login as LADOT
2. Go to Barrington project
3. Next button is disabled

Suspected cause: the initial check is not working correctly.
@KPHowley

PR #227, opened 2020-02-05
Disable forward button and show error when required fields are empty

fyliu:
Disable forward button when any required field is empty. Show error on defocusing a required field that is empty.

#218


fyliu:
- [ ] The forward button should be set initially.
- [ ] There should be an onChange trigger for the required input fields to set the error messages and forward button.

fyliu:
I updated the pull request to cover the issues found.

fyliu:
Fixes #218

PR #246, opened 2020-02-18
Check current value not empty for required rules

fyliu:
This fixes the problem of the forward button being disabled when viewing
existing projects with the required values filled in.

Fix for #218


fyliu:
Both existing projects and new projects should have the correct forward button behavior when loading the first page.

Issue #231, opened 2020-02-11
Architectural Decision: Testing frameworks


KPHowley:
### Overview
Research various testing frameworks (Cypress, Selenium, and React) to determine the best testing practices and procedures to implement. 

### Action Items
- [x]  @bradley prepare research on React Testing Library
- [x]  @nclairesays  prepare research on Cypress
- [x]  @fyliu prepare  research on Selenium
- [x] Group meeting on 2/11 to decide
- [x] Follow up meeting on 2/11 to review

### Resources/Instructions
Template:
 https://docs.google.com/document/d/1qvTyjaS0CnmT3umqWg9LJmAcZzHL6tpi5eCQWeh6ovQ/edit?usp=sharing

KPHowley:
follows up #207

fyliu:
I added what I have in the page following the template.
I also changed the file name to match this issue.

fyliu:
We discussed cypress and selenium and I felt that both are very similar in usage. The tangible difference is that selenium requires installation of webdriver binaries for target browsers, which John had issues with setting up for the Jobs-for-Hope project. Cypress likes to show the replay feature, which is potentially good in regression testing as a debugging tool. It feels less useful for acceptance testing. We've decided to implement a test case in both to compare next week.

@nclairesays will create cypress test for Barrington
@fyliu will create selenium test for Barrington

fyliu:
Continuing the discussion with what I found while working toward Selenium testing.

https://medium.com/welldone-software/an-overview-of-javascript-testing-in-2019-264e19514d0a

A couple disadvantages of cypress, side-effects of its running as a layer **inside** the browser rather than along side it:
- no headless testing (slower since it needs to render the browser to get its results)
- no drag-and-drop
Mainly Chrome support with beta support of firefox.

Selenium itself is built for browser automation and not specifically for testing although it can be used that way. The webdriver it provides is used by other tools that are made for testing, the most used one seems to be WebDriverIO.

https://www.cypress.io/how-it-works

[...]

PR #255, opened 2020-02-25
Add Cypress tests with Barrington Condos test case

nclairesays:
#231  Added basic "happy path" integration tests (aka acceptance aka journey aka end-to end) for the Barrington condos test case. Rebased off of current changes on develop and fixed merge conflicts.


nclairesays:
> 2\. 3be2c undid the WizardRuleMeasure -> RuleMeasure component name change in RuleStrategy.js. Maybe it should have been changed to RuleStrategy in the first place, like the filename.

We're no longer using the word "Measure." All the file names and component names should be using "Strategy"

nclairesays:
@fyliu I redid all my commit to using what's on develop as the head. and then cherry-picked each commit and manually resolved the issues. There are three new commits in place of the previous 4. I did a force push to this branch since it doesn''t affect anyone yet.

Issue #197, opened 2020-01-15
Implement a string minLength and maxLength validation rule for string or textarea inputs.


entrotech:
- [x] Add minStringLength and maxStringLength columns to the CalculationRule table.
- [x] Modify CalculationRule_Select* stored procedures to return the new values.
- [x] Update the rules to use a maxStringLength value of 400 for inputs of type string.
- [x] Modify client to use the rules to implement validation

ExperimentsInHonesty:
@entrotech Where are we at with this issue
1. Progress
2. Blocks
3. Availability
4. ETA

entrotech:
Proposed solution for review in PR #272

ExperimentsInHonesty:
@rosemary can you look at this paging on the dev environment to see if this is working the way you want. - here is the staging [url](https://tdm-calc-staging.herokuapp.com/) . click new project
![Screen Shot 2020-03-14 at 11 34 31 AM](https://user-images.githubusercontent.com/37763229/76688196-e90cc380-65e7-11ea-88b8-0197f6ab1706.png)

PR #272, opened 2020-03-03
197 darragh string length validation

entrotech:
This is a proposal for a signficant re-write of the client-side validation implementation that moves almost all the logic of client-side validation into the tdm-engine, rather than having a special context, effects and logic in the UI code. I would like @bfojas  and @fyliu to have a good look at this before proceeding. Maybe we can discuss at tonight's (3/3/20) meeting.

Though Issue #197 is just for string validation, the re-implementation also works for required field validation, number range validation and implements the logic that enables/disabled the forward page navigation logic.


KPHowley:
@entrotech - checking if this was reviewed, and if i can close the corresponding Issue

Issue #266, opened 2020-02-27
Add babel-eslint


fyliu:
### Overview
TdmCalculationContainer.js fails eslint right now because it is using a javascript syntax not yet recognized by eslint. Babel-eslint is the way to fix it.

Continuation of #262 . 

### Resources/Instructions
https://stackoverflow.com/questions/42701440/eslint-parsing-error-unexpected-token-error-for-assigned-fat-arrow-prop
https://stackoverflow.com/questions/34244888/how-do-i-configure-eslint-to-allow-fat-arrow-class-methods
https://medium.com/@chestozo/fixing-eslint-no-invalid-this-error-for-fat-arrow-class-methods-a56908ca8bb6

bfojas:
Eslint failure was corrected by changing the TdmCalculationContainer to a functional component. Issue is returned to the backlog, because adding babel-eslint to he project should still be considered.

ExperimentsInHonesty:
its no a priority right now.

KPHowley:
@fyliu  - Is this still needed?

fyliu:
@KPHowley It's been too long. I'm not sure. We should close it.

Issue #242, opened 2020-02-16
Make checking at least one option on page 2 of the Wizard Mandatory


KPHowley:
### Overview
Before going past we need to make sure the user selects at least one option on the "What kind of development is your project?" page so we don't display a black page or forever spinning wheel. 

### Action Items
-[] Make pressing forward disabled until an option is chosen
-[] Add "* please select at least one option"

bfojas:
Page 1 validations need to match page 2 validation. Decision is to keep the 'next' button active, but display warning instead of moving to next page when validations fail

entrotech:
@bfojas  Don't know what work has been done on this, but I provided a proposed implementation as part of PR #272.

PR #278, opened 2020-03-09
Validations and wizard routing

bfojas:
- Issue #242 . Validations are now in place for wizard page 2 and are uniform with validations of page 1.
- Wizard routing moved to react-router, allowing for easier page checking through route params rather than state.
- Refreshing/reloading while inside the wizard now reverts the user to either page 1 or page 6 depending on access control.
- "New Project" link in nav bar is now checking route rather than context.
- Issue #266 . babel-eslint was not added to the project, but linter issue that was blocking commits was fixed by converting the TdmCalculationContainer into a functional component, as per @entrotech and @fyliu suggestion. Addition of babel-eslint module will be further discussed.

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -83,7 +83,7 @@ const App = () => {
           <div className={classes.root}>
             <Route exact path="/" component={LandingPage} />
             <Route
-              path="/calculation/:page?"
+              path="/calculation/:page/:projectId?"
               render={() => (
                 <TdmCalculationContainer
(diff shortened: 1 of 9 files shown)


entrotech:
I made some modifications to fix infinite loop by separating the useEffects hooj into two different hooks, and modifying the dependency arrays for each hook to be much more precisely targeted to prevent inadvertent excess triggering of each hook.

Issue #287, opened 2020-03-15
Refactor: Fix NPM package vulnerabilities


entrotech:
Running npm i on client and server shows over 40,000 npm package vulnerabilities.  Run npm audit on both and attempt to resolve any issues.

ExperimentsInHonesty:
This issue requires someone very experienced with React

fyliu:
I made as much progress as I could at this point.
These are the remaining ones:
- 1 high severity vulnerability is not yet patched. Times 5 places in storybook that depends on it.
- 1 low severity that's patched but the newest react-scripts has not moved to the patched version of the package

PR #289, opened 2020-03-15
Bump npm versions for client and server

entrotech:
This fixed npm audit problems for the node server, but still left about 19,000 low severity vulnerabilities in the client app - ran npm audit / npm audit fix several times and it still did not get rid of these vulnerabilities, though the app and tests seem to run fine.


entrotech:
Fixes Issue #287

entrotech:
Fixes #287

PR #290, opened 2020-03-15
Fix issue 221

### LINKED ISSUE #221: Request for calculation rule
### Overview
A calculation rule to be implemented that caps the value that development can receive when two strategies are both checked.

### Action Items
See the snipit from slack below:
Discuss challenges and if we are waiting for John before we tackle.

### Resources/Instructions
snipit from slack
Rosemary McCarron  7 minutes ago
This is not an immediate concern, but Education, Marketing, Outreach, and Voluntary Travel Behavior-Change Program should be mutually exclusive because one is a less intensive version of the other. We still want the option for users to select both because Education, Marketing, Outreach is part of the packages. But if both are selected, the total earned points should be 6, not 10.

Rosemary McCarron 
(discussed with John at a Hack night)
image (7)

Bonnie  
@Rosemary McCarron thanks for the clarification.  We will work on that, but it won’t be available today.

entrotech:
Fixes  problems with enabling/disabling the Commercial Package Button related to the recent changes concerning the Education, Marketing and Outreach strategy.

Fixes #221

### DIFF
--- a/client/src/components/ProjectWizard/TdmCalculationWizard.js
+++ b/client/src/components/ProjectWizard/TdmCalculationWizard.js
@@ -85,16 +85,15 @@ const TdmCalculationWizard = props => {
     onPkgSelect,
     resultRuleCodes,
     account,
-    // projectId,
     loginId,
     onSave,
-    // onPageChange,
     onViewChange,
-    pageNo,
+    //pageNo,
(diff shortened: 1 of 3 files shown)


entrotech:
Fixes #221

Issue #307, opened 2020-03-21
When admin views project not belonging to them, they are taken to the first wizard page, and cannot navigate to other pages.


KPHowley:
### Overview

An admin should be able to open another user's project and navigage among the pages in the project, and even experiment with making changes to the inputs and strategies, but should not be able to save changes to the project.

PR #308, opened 2020-03-23
Enable nav buttons on wizard for admin user

entrotech:


entrotech:
Fixes #307 Closes #307 Resolves #307

Issue #309, opened 2020-03-23
Allow users to do CRUD operations on their own projects


entrotech:
If a user is not an admin, only allow them to see their own projects

- [x] Modify all project routes in node to extract the user's loginId from the JWT using existing middleware.
- [x] For POST request, no further change required.
- [x] For PUT request, Web API should return 403 if loginId does not match loginId of project.
- [x] For DELETE request, stored procedure should accept user's loginId as a parameter, and only delete the  project if it matches project.loginId
- [x] For getAll and getById service methods, modify sproc and service to pass user's loginId to sproc, and modify sproc to only return projects that are authored by the user (unless the user has admin role).

- [x] Hide Project Menu Item from NavBar if user is not logged in.

nclairesays:
Found a bug - you can create or delete a project with anybody's loginId via postman if you add their loginId in the request body.  (this will get closed with the next pull request) -->

nclairesays:
Reopening this issue to remove ability for admin to update and create projects.  Need to ask Rosemary if admins should be allowed to delete projects. I also think i messed up on the code logic so will fix that.

ExperimentsInHonesty:
@ramccarr @Dsomers74 
Who can Create, Read, Update or Delete (CRUD) , whose projects?

### Recommendation
#### Delete
recommendation for LADOT: If LADOT admins cannot delete users projects, then if a user complains that they put a project in and it's now gone, LADOT can confidently say, that no person has access to delete other than the user themselves.
recommendation for USER: can delete own projects

#### Update
recommendation for LADOT: If LADOT admins cannot update users projects, then if a user complains that they put a project in and it's now changed, LADOT can confidently say, that no person has access to update other than the user themselves.
recommendation for USER: can update own projects

#### Read
recommendation for LADOT: LADOT admins can read all projects
recommendation for USER: can read their own projects

#### Create:
recommendation for LADOT: LADOT admins can create their own projects
recommendation for USER: can create their own project

Do you agree with these recommendations or have questions?

KPHowley:
The above recommendations has been approved (admins should not be allowed to delete or update projects) 

@nclairesays - can you just verify that this is the current implementation in the back end, and then we can close this issue.

nclairesays:
@KPHowley yes i can do that this weekend. 

Regarding delete -- I was working on cypress tests, I discovered that the delete function has  code/logic missing so the current version does not have that update yet. I should be able to push an update sunday or monday evening.  

[...]

PR #310, opened 2020-03-23
User can only retrieve and/or edit their own projects

entrotech:
An admin user should be able to view all Projects on the Projects page.
A non-admin user should only be able to view their own projects.
Implement these rules at the Web API level.


entrotech:
Fixes #309

Issue #306, opened 2020-03-21
Add E2E test for login


fyliu:
Integration/E2E test:
- [x]  Write cypress test -- visit app, login, validate that you can see your name [Louis]
- [x]  Add cypress helper to auto login as different types of users (regular, admin, security admin) [Claire]
- [x] valid login (Cypress automatically adds cookies in the browser) [Claire]
- [x] invalid login with wrong password [Claire]

*does not include unit tests anymore*

fyliu:
I had trouble mocking async code and tried jest/enzyme and react testing library. React testing library seems to do the trick. I still need to add the actual tests. This should be done by Wednesday if no other issues come up.

nclairesays:
Cypress tests in progress. Adding edge cases for what happens during failed login attempts. Eta hopefully Wednesday also. 

@fyliu you can try mocking out async calls using the nock library

fyliu:
I checked that this works and the code looks good.

PR #333, opened 2020-04-02
Remove isAdmin helper function for Put, Post, Delete & Cypress Tests

### LINKED ISSUE #309: Allow users to do CRUD operations on their own projects

nclairesays:
First commit closes #309 

I removed ability for admin to update, delete, and create projects. The logic was also incorrect in the code because there is not an isAdmin property in the user object of the request (only in the database table). So I removed everything related to checking if they are an admin since admin should only be able to view and there seems to be logic for that already.

Second and third commit:  #306


nclairesays:
Got mixed up between what a unit test and integration should entail. I removed the more unit-y tests from cypress tests.

PR #355, opened 2020-04-30
CRUD + Cypress Tests + Bcrypt Update

### LINKED ISSUE #303: Create E2E test on project user flow
Create end-to-end (Cypress) tests around a user's workflow in regards to projects.  This entails:
1. Navigating to an existing project (after logging in)
2. Create a new project, saving it, and seeing it displayed in /projects
3. Updating an existing project

Note:
- These tests will programmatically log in (ie. does not include logging in via UI, which is covered by other tests)
- When updating an existing project, that project will be programmatically created/seeded beforehand. Test will only test the updating functionality with UI. 
- Because these tests are using a real database (ie projects will actually get created in the db), this means that created test projects should be deleted before each running test in order to clean up state/data

Blockers: 
- This issue relies on a DELETE api endpoint in order to delete data and reseed new data. (#302)

### LINKED ISSUE #309: Allow users to do CRUD operations on their own projects

nclairesays:
1. Refactor and fix CRUD functions in project controller -- Closes #309 
2. Update bcrypt 

Related to Cypress Tests:
1. Rename directory for test cases calculations
2. Fix test cases' input flow (Now selects "Education, Marketing & Outreach" as part of "Encouragement Programs")
3. Delete unused cypress files
4.  Add tests to create, read, update project logged in as a regular user -- Closes #303


app/controllers/project.controller.js
```diff
@@ -37,37 +35,42 @@ const post = async (req, res) => {
 
 const put = async (req, res) => {
   try {
-    if (!isAuthorizedUser(req)) {
+    const project = await getProject(req, res);
+    if (project.loginId !== req.user.id) {
```
fyliu, line 39: This is always going to be true since project.loginId is what's passed in as req.user.id.
Suggest referencing "const post" comparison logic above.
Suggest moving getProject call after the comparison to potentially save cycles.

app/controllers/project.controller.js
```diff
@@ -37,37 +35,42 @@ const post = async (req, res) => {
...
   } catch (err) {
     res.status(500).send(err);
   }
 };
 
 const del = async (req, res) => {
   try {
-    if (!isAuthorizedUser(req)) {
+    const project = await getProject(req, res);
+    if (project.loginId !== req.user.id) {
```
fyliu, line 54: Same comment at "const put"

app/controllers/project.controller.js
```diff
@@ -37,37 +35,42 @@ const post = async (req, res) => {
...
 
 // HELPER FUNCTION:
-
-const isAuthorizedUser = req => {
-  const loginIdOfCurrentUser = req.user.id;
-  const loginIdInRequestBody = req.body.loginId;
-
-  return loginIdOfCurrentUser === loginIdInRequestBody ? true : false;
+const getProject = async (req, res) => {
+  const project = await projectService.getById(req.user.id, req.params.id); //TODO: Check on purpose of sending req.user.id
```
fyliu, line 68: fyi. req.user.id is used by the sproc to filter for projects created by that user.

Issue #361, opened 2020-05-02
Change Text on Buttons for Single Page View and Default View


KPHowley:
### Overview
Rename the buttons to switch the views on the TDM calculation to "Switch to Single Page View"" view and "Switch to Wizard View"

### Action Items

- [x] In the wizard change button saying "Switch to default view" to "Switch to Single Page View"
- [ ] In the single page view change button saying "Switch to view" to "Switch to Wizard View"

PR #363, opened 2020-05-04
change button names, adjust button height to fit content

lemazza:
Changed Button Names to switch between Wizard and Single Page Views

adjusted button height slightly to accommodate longer names on smaller sized screens.

closes issue #361


KPHowley:
@fyliu - can you review and accept this request?

PR #370, opened 2020-05-08
Update tests after land use page removal

fyliu:
Various updates related to new changes introduced after the tests were initially merges, mainly related to removal of the land use page.


fyliu:
#309

PR #408, opened 2020-06-04
Change Caption on Summary Page to Baseline Parking

### LINKED ISSUE #404: Replace 'Req'd Parking' with 'Baseline Parking' on the TDM Calculation Summary
### Overview
Replace 'Req'd Parking' with 'Baseline Parking' on the TDM Calculation Summary

entrotech:
Closes #404 
Note: Also requires small database change - put in Database Scripts file for now, but might go into datbase migration file if we are ready.

### DIFF
--- a/.deployment
+++ b/.deployment
@@ -1,0 +1,2 @@
+[config]
+SCM_DO_BUILD_DURING_DEPLOYMENT=true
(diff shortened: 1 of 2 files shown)


entrotech:
> Hi @entrotech, what is the purpose of this new .deployment file and the content inside it? Is it relevant to changing the caption on summary page to baseline parking?

You are right. This is not relevant. I think it got added when I tried an Azure experiment.  I deleted te file from the PR and merged.   Tyler will probably set up  something different.

PR #522, opened 2020-08-13
Ui testing

croor005:


nclairesays:
Per @croor005 and my call earlier today, we will be closing this PR (no merge).

Outcome:
1) We will be continuing using cypress-testing-library for the query selectors as it is considered more symantic and more readable. The goal is so future developers and easily understand the tests. Additionally, because we are using cypress for end-to-end tests, we want to test our app from the viewpoint of a user and using the UI's content rather than css selectors to query and test our element.  Also, we will not be adding another cypress package to package.json because it is already a peer dependency to cypress-testing-library.
2) I'll be rewriting the barrington condos test and fix up some other setup configs so that @croor005 can use it as a baseline. I'll also ask about storing environment variables in github in tomorrow's meeting..
3) @croor005  will be revising the cypress.md file and submit PR without the UI cypress tests and including the CI config-related files for his next PR

Issue #539, opened 2020-08-27
Create a page for the Privacy policy and link to it on the footer


KPHowley:
### Overview
We do have a privacy policy which will be linked on the bottom of the page. The final language is still TDB but a working sheet will be put  in the resources. We need to create a page to display this.
 
### Action Items

- [x] Create page 

- [x] Create link on footer to page

jahpe777:
I created a page for "Terms and Conditions" and implemented a functional footer link. #539

![Screen Shot 2020-09-13 at 7 58 28 PM](https://user-images.githubusercontent.com/40730303/93039134-95f9e480-f5fb-11ea-8ac7-a313ed6d939a.png)

PR #561, opened 2020-09-14
Created page for Privacy Policy and implemented a functional footer link

jahpe777:
I created a page for "Terms and Conditions" and implemented a functional footer link. #539

Screen Shot 2020-09-13 at 7 58 28 PM


nclairesays (approved):
Approving so that we can have functional footer links.  I also see opportunities to make a new reusable component that the Privacy Policy, Terms and Condition, and About pages can all use since they would have the same styling. but just different children and heading.

Issue #563, opened 2020-09-14
Make buttons consistent


nclairesays:
This issue came about when working on #562 update cypress test because cypress couldn't identify which html elements were nav buttons because they were created as nested divs.  Major refactoring was identified as necessary.

Fix major button inconsistencies
 -  change buttons that are divs into actual buttons
 - make buttons reusable where possible
 - refactor the styling (there are button styles in many files that aren't related to buttons) 
 - make sure button components are responsible for itself (single responsibility)

PR #564, opened 2020-09-14
Update cypress tests & refactor buttons

### LINKED ISSUE #562: Update cypress tests
Update cypress tests so that we can run at least one in our pipeline.  It's been a while since the tests have been updated to reflect how users interact with the app, e.g. new form requirements, updated strategies, etc.

nclairesays:
Closes #562 and #563 

Cypress related changes (#562):
- Delete unused cypress example tests
- Refactor cypress tests and NavButton; Barrington Condos passing
- Refactor Level0-related page and styling from ProjectTargetPoints into its own component

Buttons related changes (#563):
- Add Button component; LOTS of refactoring and styling changes
-- Add jss theme and provider
-- Refactor NavButtons to make it more readable
-- Moved nav logic from TdmCalculationWizard into the NavButton component
-- Create new Button directory and move button components in
-- Remove unused button styling from App.css and Calculation.scss
-- Clean up/refactor code in TdmCalculationWizard, Level0Page, and ProjectTargetPoints
- Update single page view to use Button component to switch views


nclairesays:
How the pages look like. I changed the "switch view" button to all caps like the other buttons, and felt the caps was enough emphasis to indicate that it's a button. If designers want to change it back to green or a different color, we can do that pretty easily now. 

image
image
image
image

nclairesays:
Single Page view uses the same Button component 
image

entrotech (approved):
Wow, that's a lot of changes to put in one PR. It would be better to have any re-factoring changes in separate PRs and any changes to cypress tests in separate PRs, to make it easier to manage the review, merge and deployment.  There seem to be a number of changes to the button behavior - for example, it now changes appearance when it has focus to have a border with rounded corners when it has focus (whereas it has square corners and no border when it doesn't have focus). Also the drop-shadow is gone on the nav buttons, and the <, > symbols are much bolder.  Back when we had other UI/UX people, the square corners, lighter font, etc. were points that they specifially requested.  On the other hand, using real buttons follows UI conventions, in that you can navigate the UI from the keyboard, since you can "press" the button with focus by tabbing to it and pressing the <Enter> key.  We'll see what LADOT has to say today.

nclairesays:
@entrotech sorry about the big PR. Fixing the cypress tests meant fixing the buttons so that the tests can find the button. I initially was thinking that my commits would be a good separator the changes, but looking back, i can see how they could have been in separate PRs. 

As for the button behaviors, I didn't add any new behaviors; rather whatever behavior is there, is how the native button html component works. I also just copied and pasted whatever styling for the nav button i found. 

Regarding the drop shadow, I dug into it just now -- it looks like there were actually two `.tdm-wizard-nav-button`  objects in the Calculation.scss file. And I happened to copy the one that did not have a drop shadow. We can add that back if designers and LADOT think it is best.

Issue #572, opened 2020-09-19
Create a pop up box or "Modal" for the Terms and Conditions 


KPHowley:
### Overview
We need users to accept a terms and conditions every time they visit the Calculator and so we need a pop up modal requiring them to do so. 

### Action Items

- [x] Create a pop-up box or modal in accordance with the design in resources. 
- [x] When users press the Decline button to redirect them back to the LA DOT home page (https://ladot.lacity.org/)
- [x] The box should appear every time the user first comes to the site, through whatever page, however, acceptance can be stored in a local cookie, only for the current session. 

### Resources/Instructions
![93415316-161c8600-f858-11ea-8d97-10f37336c5d7](https://user-images.githubusercontent.com/49739276/93688144-873a8400-fa78-11ea-9a33-04c567194dc5.png)

jahpe777:
I created a modal in accordance with the design in resources. When users click the Decline button it redirects them back to the LA DOT home page (https://ladot.lacity.org/). The modal appears when the user first comes to the site. After accepting the "Terms & Conditions", the user is authorized (via LocalStorage) to use the calculator and will no longer be required to accept the "Terms and Conditions" again unless LocalStorage is cleared.

KPHowley:
Hi James, @jahpe777  - This looks good, and that should work as far as redirect and storing acceptance info. Can you make sure the text displayed is the following, and that users can scroll to view it all: 

TDM Calculator User Terms and Conditions 

PLEASE READ THIS AGREEMENT CAREFULLY BEFORE USING THIS WEB SITE. BY USING THIS WEB SITE, YOU ARE CONSENTING TO BE OBLIGATED AND BECOME A PARTY TO THIS AGREEMENT. IF YOU DO NOT AGREE TO THE TERMS AND CONDITIONS BELOW YOU SHOULD NOT ACCESS OR USE THIS WEB SITE
[...]

PR #594, opened 2020-10-11
Create a pop up box or "Modal" for the Terms and Conditions #572

jahpe777:
#572

I created a modal in accordance with the design in resources. When users click the Decline button it redirects them back to the LA DOT home page (https://ladot.lacity.org/). The modal appears when the user first comes to the site. After accepting the "Terms & Conditions", the user is authorized (via LocalStorage) to use the calculator and will no longer be required to accept the "Terms and Conditions" again unless LocalStorage is cleared.

Screen Shot 2020-10-11 at 3 57 19 PM


nclairesays:
@jahpe777 Can you move everything related to the terms and conditions modal  into a new file/component and then import that component into the the TdmCalculationWizard.js? (e.g you can export/import the toggleTermsAndConditionsModal, and you might have to pass in props). The useState should also be inside that new file. We should be abiding to the single responsibility rule and not try to overload one component (e.g. TdmCalculationWizard) with all the terms and conditions code/styling. 

Secondly, can you try using the new reusable Button.js component rather than creating a new one with its own styling? We should try to not have duplication of code-- which is why I went down the road of creating the reusable Button components. In this case, there is an opportunity here to keep consistent and use it here! Feel free to modify the Button if needed, but make sure the other places that use that Button still work as expected.

client/src/components/TermsAndConditionsModal.js
```diff
@@ -0,0 +1,308 @@
...
+          to confirm the validity of the data provided.
+        </p>
+      </div>
+
+      <div className={classes.modalActions}>
+        <Button
+          className={classes.declineBtn}
+          onClick={e => {
+            e.preventDefault();
+            window.location.href = "https://ladot.lacity.org/";
```
nclairesays, line 283: Can you use `history.push("https://ladot.lacity.org/") here instead of window.location?

nclairesays (approved):
Approving changes. There are a couple things that need to be changed a bit but we can discuss that during our weekly meetings.

client/src/components/TermsAndConditionsModal.js
```diff
@@ -0,0 +1,308 @@
...
+    },
+    "& input": {
+      boxSizing: "border-box",
+      fontSize: "20px",
+      lineHeight: "24px",
+      padding: "16px",
+      border: "1px solid #979797",
+      marginTop: "8px"
+    },
+    "& .scroll": {
```
nclairesays, line 43: this should be its own key and not nested like this. you'd then use in your code by calling classes.scroll

client/src/components/TermsAndConditionsModal.js
```diff
@@ -0,0 +1,308 @@
...
+    },
+    "& p": {
+      fontSize: "16px",
+      lineHeight: "25px",
+      textAlign: "left",
+      "& img": {
+        margin: "4px 12px 12px 0"
+      }
+    },
+    "& .smaller": {
```
nclairesays, line 30: `smaller` should be a key name and not nested under modal. you can use it in your code by calling classes.smaller as the classname

client/src/components/TermsAndConditionsModal.js
```diff
@@ -0,0 +1,308 @@
...
+    <Modal
+      isOpen={termsAndConditionsModalOpen}
+      onRequestClose={toggleTermsAndConditionsModal}
+      shouldCloseOnOverlayClick={false}
+      contentLabel="Terms and Conditions Modal"
+      style={modalStyles}
+      className={classes.modal}
+    >
+      <h2>TDM Calculator User Terms and Conditions</h2>
+      <div className="scroll">
```
nclairesays, line 144: classname should be {classes.scroll} once you put `scroll` as the key in the useStyles

client/src/components/TermsAndConditionsModal.js
```diff
@@ -0,0 +1,308 @@
...
+        <p>
+          PLEASE READ THIS AGREEMENT CAREFULLY BEFORE USING THIS WEB SITE. BY
+          USING THIS WEB SITE, YOU ARE CONSENTING TO BE OBLIGATED AND BECOME A
+          PARTY TO THIS AGREEMENT. IF YOU DO NOT AGREE TO THE TERMS AND
+          CONDITIONS BEFLOW YOU SHOULD NOT ACCESS OR USE THIS WEB SITE
+        </p>
+
+        <p className="smaller">
```
nclairesays, line 152: should be using {classes.smaller} once you fix it in the usestyles

client/src/components/TermsAndConditionsModal.js
jahpe777: Hi Claire, I tried subbing out window.location with history.push() but it just added on to the end of the URL. Is there like a history.clear() method that will reset the URL and push the new one?

Screen Shot 2020-10-25 at 2 50 53 PM

nclairesays: @jahpe777 i see. i don't have an immediate solution for this. just use window.location since we know that works for navigating outside our site. don't want you to spend too much time on this.

jahpe777: Ok, thanks!

PR #600, opened 2020-10-17
349 Login Alert Tooltip

Jonathanko52:
Create a tooltip that pops in under the login button to remind users to login in to TDM if they wish to save their projects.


client/src/components/NavBarLogin.js
```diff
@@ -41,7 +58,14 @@ const NavBarLogin = ({
   );
 
   return !account || !account.email ? (
-    loginLink
+    !isCalculation ? (
+      <React.Fragment>{loginLink}</React.Fragment>
+    ) : (
+      <React.Fragment>
+        <NavBarToolTip></NavBarToolTip>
```
nclairesays, line 65: This can be a self closing component

nclairesays: I resolved this with my last commit

Issue #506, opened 2020-08-01
Bug: The toast to notify users after login information has timed out is displayed twice


KPHowley:
### Overview
After a user's is logged out due to having their security authentication time out, the toast which notifies them ids displayed twice. It should only be displayed once. 

### Action Items

- [x] Further document the issue 
- [x] Bug fix by front end(?)

KPHowley:
@jahpe777  give us an update on:
    1. Progress
    2. Blockers
    3. Availability
    4. ETA

jahpe777:
Edited the toast to only display once in order to notify users after login information has timed out. #506

![1](https://user-images.githubusercontent.com/40730303/97125143-27c33a00-16f0-11eb-83f5-9e00701ddf31.jpg)

PR #610, opened 2020-10-26
Bug: The toast to notify users after login information has timed out is displayed twice #506 

jahpe777:
Edited the toast to only display once in order to notify users after login information has timed out. #506

Also included styling changes to scrollbar for the Terms and Conditions Modal per request.

### DIFF
--- a/client/src/components/Projects.js
+++ b/client/src/components/Projects.js
@@ -261,7 +261,7 @@ const Projects = ({ account, history }) => {
       }
     };
     getProjects();
-  }, [email, toastAdd, historyPush]);
+  }, [email, historyPush]);
 
   const toggleDuplicateModal = async project => {
     if (project) {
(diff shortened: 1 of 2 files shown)


nclairesays (approved):

KPHowley:
@jahpe777 @nclairesays  - Hey I was just reviewing this and I saw a couple things I wanted to discuss about the modal 

 First, I see the popup only comes up on /calculations, which I think is fine, but i believe is a changes.  Also the pop up seems to be pinned to the bottom of my window, and not in the center, can we make sure it's centered? Finally we're you trying to make the scroll bar always visible -  it isn't visible for me. 

I attached a video of this on my computer as well:

https://drive.google.com/file/d/1gXW1EXDy4aAmoetWMKhrmRc8M4yEtdRk/view?usp=sharing

PR #634, opened 2020-11-07
Issue 628 Save button in TDM wizard only shows if user is logged in

Jonathanko52:
Save button only appears now if user is logged in. TDMWizard checks for account email value to verify


entrotech (changes requested):
The request in the Issue is not correct. In order for the Save button to be enabled:
If it is a new project (i.e., projectId  is falsey), then the user must be logged in.
If it is an existing project (i.e., projectId > 0), then the logged in user (account.id) must equal the project author's loginId - that is, the user must own the project.

entrotech:
I'm closing this without merging, in favor of PR #642, which implements more complete rules for when the Save button should be visible.

PR #635, opened 2020-11-07
565 save button update timestamp keep page

Jonathanko52:
Timestamp at bottom of wizard updating on save(after a one second delay for the database to update). See notes on issue for progress on the rest of the issue.

### DIFF
--- a/client/src/components/ProjectWizard/TdmCalculationWizard.js
+++ b/client/src/components/ProjectWizard/TdmCalculationWizard.js
@@ -136,21 +136,22 @@ const TdmCalculationWizard = props => {
   } = props;
   const [dateModified, setDateModified] = useState("");
 
-  useEffect(() => {
-    const getDateModified = async () => {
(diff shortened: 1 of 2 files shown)


entrotech (commented):
Sorry for this confusing issue description.

The effect and web api call to get the dateModified aren't necessary, as the project alredy has the dateModified on it.  I'm submitting an alternative implementation that passes the dateModified down from TmCalculationContainer to the wizard for display.

The logic for the Save button described in the issue is overly simplistic.  In order for the Save button to be enabled, the following must be true:
- The form is "dirty"
- The user is logged in.
- The logged in user must own the project
- There cannot be any validation errors on the project.

I'm going to close this PR without merging and replace it with a different PR.

entrotech:
I'm closing this without merging, in favor of PR #642, which involves less code and fewer server requests, since the client already has the dateModified information.

Issue #609, opened 2020-10-23
Issues while deleting files under "My Project"


VJ1285:
### Overview
When the user hits the delete button for a project under "My Projects" the deleted project still shows under the list of projects. It only disappears when the user refreshes the page (see video below)

### Resources/Instructions
[Recording #5.zip](https://github.com/hackforla/tdm-calculator/files/5432026/Recording.5.zip)

PR #648, opened 2020-11-13
#609 project deletion renders without page refresh

tan-nate:


nclairesays (changes requested):
- I like that you moved the getProjects function out of the useEffect
- Requesting changes -- it should not be the deleteProject function's responsibility to getProjects after it's deleted. When we delete a project, the "projects" array which contains all the projects on the page is updated, resulting in a state change. Same goes with the "selectedProject."  Thus, the better approach to this issue would be to add into "projects" or "selectedProject" to the useEffect dependencies array, though I'd probably with with "projects"

client/src/components/Projects.js
```diff
@@ -238,28 +238,29 @@ const Projects = ({ account, history }) => {
...
+        historyPush(`/login/${encodeURIComponent(email)}`);
       }
-    };
+      console.error(err);
+    }
+  };
+
+  useEffect(() => {
     getProjects();
   }, [email, historyPush]);
```
nclairesays, line 265: add dependency here

tan-nate:
Okay that makes sense. Shouldn't the deleteProjects function also wait to catch a possible error, and only update the array if the request is successful? Kind of like getProjects:

```
const getProjects = async () => {
    try {
      const result = await projectService.get();
      if (result.data === "" || result.data === false) {
        setProjects([]);
      } else {
        setProjects(result.data);
      }
    } catch (err) {
      // If user's session token has expired or they are not
      // authorized for this web api request, let them know
      // and redirect to login
      if (err.response && err.response.status === 401) {
        toastAdd(
          "For your security, your session has expired. Please log in again."
        );
        historyPush(`/login/${encodeURIComponent(email)}`);
      }
      console.error(err);
    }
  };
```

nclairesays:
@tan-nate yea, great call. I think having an error handler for delete is a good idea if it’s not there already.  Likely, you can do a check if something is not a 200/201/success status code. Feel free to add that.

nclairesays (commented):
Check out my comments for your information. There are a few minor things that needed to change but I'll just do it on my end because there were a couple things I was going to change this component anyway.

PR #650, opened 2020-11-18
449 add scroll to project table

Jonathanko52:
Created a container for the table, added min width for table, and has container scroll when that min width is reached, generating a scroll effect when the window is to small to comfortable fit the table.


nclairesays (changes requested):
I see the responsiveness and scrollbar. However, the width of the table seems to be smaller now. We still want the table to optimize as much space as possible so the size of the table should stay the same as it was before if/when they are viewing the website on a full/normal screen. It's when the screen size gets smaller, we want it to be responsive/fit accordingly. 

Also, the edge of the search bar and the table don't seem to be aligned anymore. Compare with original table image below. 

Maybe to start with a baseline, you can put your chrome window to 1920px by 1080px which is a very common screen size. Then from there, you can make your screen smaller and see how it responds and make sure users have a way to see the content. 

Let me know if you have questions.

image

image

PR #661, opened 2020-11-26
Fix nav and login bug related to setting account and local storage #660

### LINKED ISSUE #660: Fix bug related to setting account in app and currentUser in local storage
### Overview
Two features are affected by this bug:
1. Users cannot navigate past the first page of /calculation when in incognito due to localStorage's currentUser being null
2. NavBar does not recognize that user is logged in even tho there is a currentUser in local storage (Shows "Create Project" and "Login" button when it should show "My Projects" and "Logout")

nclairesays:
Closes #660 

- Made sure to set local storage to `{}` if there is no currentUser in localStorage
- Refactored and removed unnecessary setAccount in NavBarLogin component


entrotech (approved):
Cool. I like PRs that remove unnecessary code!

PR #690, opened 2020-12-18
Toast now centered on right side container

Jonathanko52:
Toast is now centered on the right side of the page, ignoring the sidebar, and has Z index set to one so it pops up over the sidebar in mobile view.


entrotech (approved):
Do you really want toast duration to be 555 days?

Issue #671, opened 2020-12-03
Refactor: Tool tips


KPHowley:
### Overview

Some of the tool tips are different on the page, we should refactor them to be consistent.

Jonathanko52:
Covered in #407, with pull request https://github.com/hackforla/tdm-calculator/pull/681

Jonathanko52:
-Standardize sidebar and calculator strategies tooltips.
-Both move with mouse.
-Both have arrows.
-Darker blue test (same as current sidebar.)
-Border(same as strategies)
-Background white(same as strategies)

Jonathanko52:
Narrowed down that the sidebar uses tooltip.js(from the sidebar field) while the wizard uses react tooltips npm package. Should be a simple fix from here. Eta 1/7

Jonathanko52:
Pull request submitted

Jonathanko52:
Pull request submitted. Awaiting review.

nclairesays:
I approving JK's PR but this issue isn't closed yet. I'll work to continue refactoring this. 

There are some files that could be deleted, some styling that's passed into a component when the component should already be responsible for its own styling. There's a lot of repetitive code in these tool tips so I will work to refactor and clean them up.

nclairesays:
Still need to fix the tool tip that explains the packages when you're on level 1. The ProjectsPackages.js uses the old tool t ip
![image](https://user-images.githubusercontent.com/45305029/105144329-ec979380-5ab1-11eb-859a-7a2a3f2184a2.png)

nclairesays:
I did some additional refactoring on top of what JK already did (which has already been approved and merged in)

Mostly reorganizing files/components so that they are DRY
- Created a Tooltip directory with ToolTipIcon.js and ToolTip.js. (This ToolTip.js is _different_ from our previous versions of ToolTip.js) 
- These are the components we should be importing throughout the app, rather than importing directly from the library. 
- These two components are responsible for their own styling. Styling does _not_ get passed into the component anymore.
- Made all ToolTipIcons the same, with the option of making it a "medium" size via a prop

Bug FIx: Fixed the rendering of tooltip icon for Parking Provided / Baseline and City wide Parking baseline since the icon accidentally got added in

PR #699, opened 2021-01-08
#671 Refactored Sidebar and Calculator Tooltips

Jonathanko52:
Changed both tooltips to the react-tooltips library and standardized the CSS. However, have not removed the old Sidebar/tooltip file. 

#671

### DIFF
--- a/client/src/components/ProjectWizard/RuleCalculation/RuleCalculation.js
+++ b/client/src/components/ProjectWizard/RuleCalculation/RuleCalculation.js
@@ -1,8 +1,9 @@
 import React, { useState } from "react";
 import PropTypes from "prop-types";
 import { createUseStyles } from "react-jss";
-import ToolTip from "./ToolTip";
+import ToolTipIcon from "./ToolTipIcon";
 import clsx from "clsx";
+import ReactTooltip from "react-tooltip";
 
 const useStyles = createUseStyles({
   field: {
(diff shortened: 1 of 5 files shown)


client/src/components/ProjectWizard/RuleCalculation/RuleCalculation.js
```diff
@@ -325,7 +345,13 @@ const RuleCalculation = ({
             data-class={classes.tooltip}
           >
             {name}
-            <ToolTip tipText={description} />
+            <ToolTipIcon
```
nclairesays, line 348: image
There shouldn't be a tooltip for the parking provided/baseline.  It should only show a tooltip IF there's a description.  So since the bottom doesn't have a description, the icon should not be showing.

client/src/components/ProjectWizard/RuleCalculation/RuleCalculation.js
```diff
@@ -147,6 +148,25 @@ const useStyles = createUseStyles({
...
+  },
+  tooltip: {
+    color: "rgb(30, 36, 63) !important",
+    minWidth: "200px",
+    maxWidth: "400px",
+    fontFamily: "Arial",
+    fontSize: 12,
+    lineHeight: "16px",
+    fontWeight: "bold",
+    "-webkit-box-shadow": "0px 0px 8px rgba(0, 46, 109, 0.2)",
```
nclairesays, line 160: i don't think you need add the webkit and moz. is there a reason why you was these? i think most modern browsers support box shadow and definitely support border radius

nclairesays: I'm assuming this may have been a copy pasta? but if there's a reason for adding them, definitely open to hearing why.

nclairesays (approved):
I'm approving this PR but this issue isn't closed yet. I'll work to continue refactoring this. 

There are some files that could be deleted, some styling that's passed into a component when the component should already be responsible for its own styling. There's a lot of repetitive code in these tool tips so I will work to refactor and clean them up.

PR #704, opened 2021-01-14
470 sidebar ref solution

Jonathanko52:
Set the toast's createPortal target to the TDM Wizard/Main container components through the use of refs, automatically centering the toasts to these components. This means the Toast is now centered to these targets rather than the document.body.

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -34,14 +34,19 @@ const App = ({
   account,
   setLoggedInAccount,
   hasConfirmedTransition,
-  sidebarRef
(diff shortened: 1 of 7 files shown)


entrotech (approved):
Nice work!  Seems like a lot of effort just to move the toast location a bit, but I like how it turned out.

entrotech:
I did have to resolve merge conflicts with the develop branch on the App.js file. Most likely, this means that you did not merge the develop branch into your feature branch before pushing in to github.

PR #720, opened 2021-01-20
671 refactor tooltips pt2

nclairesays:
I did some additional refactoring on top of what JK already did (which has already been approved and merged in) for issue #671 

For whoever is reviewing my PR, it might help to review one commit at a time. I tried to make my commit name descriptive. Click on the 3 dots below to expand.

Mostly reorganizing files/components so that they are DRY
- Created a Tooltip directory with ToolTipIcon.js and ToolTip.js. (This ToolTip.js is _different_ from our previous versions of ToolTip.js) 
- These are the components we should be importing throughout the app, rather than importing directly from the library. 
- These two components are responsible for their own styling. Styling does _not_ get passed into the component anymore.
- Made all ToolTipIcons the same, with the option of making it a "medium" size via a prop

Bug FIx: Now conditionally rendering of tooltip icon for Parking Provided / Baseline and City wide Parking baseline since the icon accidentally got added in with recent PR. It now conditionally renders and aligns correctly even if it doesn't have the icon.
image


nclairesays:
sdf

Jonathanko52 (changes requested):
Screen Shot 2021-01-20 at 8 26 43 PM (2)
Screen Shot 2021-01-20 at 8 26 45 PM (2)

-Tooltip text on page 4 of the wizard should be the same as the color of the text on sidebar.

nclairesays:
@Jonathanko52 fixed page 4, as well as made sure ALL of the tooltips have the exact same styling. some of them were missing padding. I actually made some additional code changes to refactor some more but i'll make a third PR so that this one isn't overloaded. 
image

PR #728, opened 2021-01-26
722 Button Uniformity

### LINKED ISSUE #722: Make rest of buttons consistent with app (login, registration, forgot password)
All the buttons on the Calculator should use the same button type and look visually similar. currently we different buttons on the login page than the wizard. 

- [x] Update all buttons on login Page to be same color and style as "save" button on the wizard pages. 
- [x] Update forgot password buttons
- [x] Verify all buttons elsewhere are using the same type of button class

tan-nate:
Buttons are updated to use the Button component, except for a few cases such as the "Reset Page" button in the wizard, which is left as a `<button>` to preserve styling

closes #722


client/src/components/Authorization/Login.js
```diff
@@ -192,20 +193,20 @@ const Login = props => {
                     </Link>
                   </div>
 
-                  <button
+                  <Button
                     id="cy-login-submit"
                     type="submit"
                     className="btn-submit"
```
nclairesays, line 199: We should be getting rid of classname and using the default style of the Button component

client/src/components/Authorization/Login.js
```diff
@@ -192,20 +193,20 @@ const Login = props => {
...
+                    color="colorPrimary"
                   >
                     {isSubmitting ? "Please wait..." : "Login"}
-                  </button>
-
+                  </Button>
                   {/* <button className="btn-without-saving"> */}
                   <Link to="/calculation/1">
-                    <button className="btn-without-saving">
+                    <Button className="btn-without-saving">
```
nclairesays, line 207: should get rid of css classname and use styles from the Button component props to determine the style

client/src/styles/AuthForms.scss
```diff
@@ -48,23 +48,25 @@
   }
 
   .btn-submit {
```
nclairesays, line 50: we should be able to get rid any css related to buttons

client/src/styles/AuthForms.scss
```diff
@@ -48,23 +48,25 @@
...
+    // @extend %button-shared;
     
     line-height: 1.5;
     width: 100%;
     padding: .75em;
     cursor: pointer;
+    margin: 0.5em 0;
   }
 
   .btn-without-saving {
```
nclairesays, line 60: we should be able to get rid any css related to buttons

client/src/components/Authorization/Register.js
```diff
@@ -190,13 +191,13 @@ const Register = props => {
                         />
                       </div>
 
-                      <button
+                      <Button
                         type="submit"
                         className="btn-submit"
```
nclairesays, line 196: get rid and use props

client/src/components/Authorization/Login.js
```diff
@@ -192,20 +193,20 @@ const Login = props => {
...
                     type="submit"
                     className="btn-submit"
                     disabled={isSubmitting}
+                    color="colorPrimary"
                   >
                     {isSubmitting ? "Please wait..." : "Login"}
-                  </button>
-
+                  </Button>
                   {/* <button className="btn-without-saving"> */}
```
nclairesays, line 205: should be able to delete this comment

nclairesays (changes requested):
Most of my comments say the same thing: 
- we should be able to get rid of all css related to buttons and use the jss styles that are in the Button component. You can use those styles by passing in props (i.e. variant and color). The size of the buttons do not need to be the same size as they currently are. (e.g. the login button will be a lot smaller than what it is now. Maybe use your best judgement how the login and continue without logging in buttons are laid out and we can discuss after.)
- Any "submit" type buttons should be using colorPrimary (the green color) when submitting.

nclairesays:
I made a commit ( 9c841a2 ) that cleaned up all the auth related pages so that the buttons are using the shared Button component and its related props. The styling is more consistent with the rest of the app.  

Additionally, I cleaned up the login/register component's styling and removed/moved as much css/scss styling into the their components as jss where I saw fit.  

I also fixed the hovering effect that warns the user if they are about to click on the "continue without saving" button. This was previously but somehow, the hovering/warning effect got disappeared.  The user will now get a warning that tells them that it's recommended to login.

The confirm page still needs to be properly styled. Covered by issue #738.

image

image

image

image


image

Issue #713, opened 2021-01-18
Change Colors of Earned Points and Target Points on Sidebar


KPHowley:
### Overview
Change the Color of Earned Points to be orange before equaling target points, and green once it equals or is greater than target points.  Change  Target Points to be Green on Sidebar

### Action Items

- [x] Change the Color of Earned Points to be orange before equaling target points, 
- [x] Change the Color of Earned Points to be green once it equals or is greater than target points.
- [x] Change Target Points to be Green on Sidebar

### Resources/Instructions
Figma: TBD 

![Screen Shot 2021-01-17 at 6 23 43 PM](https://user-images.githubusercontent.com/49739276/104866275-ffb03500-58f2-11eb-9c9f-63937050d6ba.png)

KPHowley:
@dreams0ng - Can you update this with the Figma link and a picture for this issues (2 of 3 issues)

dreams0ng:
1. Progress
- Updated UI to match live site
2. Blockers
- N/A
3. Availability
- Any
4. ETA
- Presented

MOCKUPS

1. ![CC1](https://user-images.githubusercontent.com/22779768/105244892-007cdd00-5b26-11eb-8482-5451a1560cdc.png)
2. ![CC2](https://user-images.githubusercontent.com/22779768/105244902-0246a080-5b26-11eb-8d02-d3bd41ab877d.png)
3. ![CC3](https://user-images.githubusercontent.com/22779768/105244921-0a9edb80-5b26-11eb-8323-547c8e438473.png)
4. ![CC4](https://user-images.githubusercontent.com/22779768/105244925-0bd00880-5b26-11eb-9d02-d36fd89f6701.png)

![CC](https://user-images.githubusercontent.com/22779768/105244943-12f71680-5b26-11eb-94ba-b3e4997c54b7.gif)

LINKS
https://www.figma.com/proto/6h72XLxK76m4INtsSu8Mt0/TDM-Calculator-(Copy)?node-id=2392%3A222&viewport=-1041%2C821%2C0.1437286138534546&scaling=min-zoom

jahpe777:
Changed Colors of Earned Points and Target Points on Sidebar #713

![Screen Shot 2021-01-26 at 11 36 09 AM](https://user-images.githubusercontent.com/40730303/105895971-11689b00-5fcb-11eb-8205-aa21ec754ce0.png)

[...]

PR #729, opened 2021-01-26
713 change color of points

jahpe777:
Changed Colors of Earned Points and Target Points on Sidebar #713
Also, flipped position of Earned and Target Points

Screen Shot 2021-01-26 at 11 36 09 AM

Screen Shot 2021-01-26 at 11 36 31 AM

Screen Shot 2021-01-26 at 11 36 48 AM

Screen Shot 2021-01-26 at 11 37 57 AM

Screen Shot 2021-01-26 at 11 38 03 AM

Screen Shot 2021-01-26 at 11 38 06 AM

Screen Shot 2021-01-26 at 11 38 19 AM


client/src/components/ProjectWizard/SidebarPoints/SidebarPoints.js
```diff
@@ -45,16 +53,30 @@ const useStyles = createUseStyles({
...
   const noToolTip = rule.value === 0 ? classes.noDisplay : "";
 
+  let target = rulesConfig.target.value;
+  let earned = rulesConfig.earned.value;
+
+  let earnedPointsColor =
+    (rule.name === "Earned Points" && earned === 0) ||
+    (rule.name === "Earned Points" && earned < target)
+      ? classes.ruleValueEarned
+      : classes.ruleValue;
```
nclairesays, line 69: So just fyi, the reason I ended up writing the above comment about the theme colors was because I didn't know what "ruleValue" was and what color that was supposed to be when reading this line. Maybe we can also rename this class to make it more descriptive. (Though I can't think of one right now.. so if you can't either then it's okay.)

client/src/components/ProjectWizard/SidebarPoints/SidebarPointsPanel.js
```diff
@@ -15,6 +15,16 @@ const SidebarPointsPanel = props => {
     earnedPointsRule = rules.filter(rule => rule.code === "PTS_EARNED")[0];
     projectLevelRule = rules.filter(rule => rule.code === "PROJECT_LEVEL")[0];
   }
+
+  const rulesConfig = {
```
nclairesays, line 19: Yep, this is a great way to get both the target and earned values in one object!

nclairesays (changes requested):
Great work. You need to change the lets to const. I'm also going to challenge you by requesting that you add/use the theme colors. See comments below.

client/src/components/ProjectWizard/SidebarPoints/SidebarPoints.js
```diff
@@ -45,16 +53,30 @@ const useStyles = createUseStyles({
...
 const SidebarPoints = props => {
   const classes = useStyles();
-  const { rule } = props;
+  const { rule, rulesConfig } = props;
+
   const opacityTest =
     rule.value && rule.value !== "0" ? "" : classes.lowOpacity;
   const noToolTip = rule.value === 0 ? classes.noDisplay : "";
 
+  let target = rulesConfig.target.value;
```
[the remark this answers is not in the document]
nclairesays: I only wrote it for this one line, but all your variables should be using const

KPHowley:
@nclairesays  - Is this just pending a review from @entrotech  or do we need any updates from @jahpe777 ?

nclairesays:
@KPHowley - I believe @jahpe777 is working on changing a few things I requested. James, let me know how you're doing on that and maybe an eta. 

We could technically approve this PR and James can make another PR with the additional updates if we want to get something in before Tuesday

nclairesays:
@jahpe777 Since there's a meeting today, I approved the PR so that they can showcase your work. Can you make a new PR with the changes I requested?

PR #735, opened 2021-02-03
Refactor and clean up Terms and Conditions page and modal #723

### LINKED ISSUE #721: Bug: Terms and condition should be center of page
Issue: When app is in a smaller width, the modal is not centered of the page. The modal also gets cut off. 

image

### LINKED ISSUE #723: Refactor -  Make Terms and Conditions Modal and Page use the same component

nclairesays:
closes #721 and closes #723 

Refactor Terms and Conditions and fix modal responsiveness 
- Create separate `TermsAndConditionsPage` and `TermsAndConditionsModal` components that use the same `TermsAndConditionsContent` component (in case this helps you read/understand the git diffs: I copied/used part of the original `TermsAndCondition` as the base for the `TermsAndConditionsContent` component, then renamed `TermsAndConditions` to `TermsAndConditionsPage`. Then imported `TermsAndConditionsContent` into `TermsAndConditionsPage` and `TermsAndConditionsModal`)
- Restyled the components to be more consistent with each other
- Fix Buttons in modal so that it uses the `Button` props (variant/color) instead of custom class
- Remove unused or duplicate styles
- Made sure it looks as expected on both Chrome and Firefox
- Made sure it's responsive

Things that are slightly different from the original:
- Made text alignment justified 
- In modal, you now have to scroll to the bottom of the modal in order to click on the Accept or Decline buttons. I initially attempted to have the scroll bar inside just the content but it was very awkward to have in the code. Also, I think this is a better user flow because it forces users to at least scroll and "read" the terms before accepting. 
- modal also covers about 80% of the page regardless of screen size

image
image

- the terms and conditions page style is also different in the width of the content is wider and page will behave like a normal page that scrolls. 

image
image


nclairesays:
Per discussion with Kevin, we decided to move the modal buttons to the right 
image

PR #745, opened 2021-02-06
713 change color of points

jahpe777:
Refactored ticket based on Claire's comments.


nclairesays:
Great job, I just added another commit to remove the "color" argument that was passed into the SidebarPoints component and the theme since it isn't getting used.

PR #764, opened 2021-03-01
#756 Fix error message and resends email when user tries to login with unconfirmed email

### LINKED ISSUE #756: Resend login confirmation email does not send email to existing accounts
### Overview
Currently, if you try to login and have not confirmed your email address, our system will resend a confirmation email and this is the message that appears:
image

It does not seem to be re-sending the confirmation email.
### Action Items
Update the error message to be more generic, since in this use case, we have registered an account. Plus, there is already a different error message for when an account does not exist yet.

Fix the bug to make sure that email is sent correctly.

### Resources/Instructions

after some  initial debugging, i've narrowed down the bug to the `resendConfirmationEmail` in the account.service.js file. it is somehow ending up in the catch block. `err.message` is `Cannot read property '0' of undefined"` when making that request.

nclairesays:
Closes #756 

Side note - Reposting from [slack post](https://hackforla.slack.com/archives/CKY65G266/p1614568234007300). It has not yet been made into an issue: 
i just tried logging in with an unconfirmed email 3 times, saw 3 registration confirmation emails, tried clicking on the oldest email link and it verified me. Not sure if this is expected behavior or if only the most recent email link should be allowed to verify the user.

### DIFF
--- a/server/app/services/account.service.js
+++ b/server/app/services/account.service.js
@@ -99,7 +99,7 @@ const resendConfirmationEmail = async email => {
     result = {
       success: true,
       code: "REG_SUCCESS",
-      newId: selectByEmailResponse.recordset.rows[0].id,
+      newId: selectByEmailResponse.recordset[0].id,
       message: "Account found."
     };
     result = await requestRegistrationConfirmation(email, result);
@@ -110,7 +110,7 @@ const resendConfirmationEmail = async email => {
     return {
       success: false,
       code: "REG_ACCOUNT_NOT_FOUND",
-      message: `Email ${email} is not registered. `
+      message: `Resending confirmation email to ${email} failed due to: ${err.message}`
     };
   }
 };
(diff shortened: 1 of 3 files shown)


server/app/services/account.service.js
```diff
@@ -110,7 +110,7 @@ const resendConfirmationEmail = async email => {
     return {
       success: false,
       code: "REG_ACCOUNT_NOT_FOUND",
-      message: `Email ${email} is not registered. `
+      message: `Resending confirmation email to ${email} failed due to: ${err.message}`
```
entrotech, line 113: You could get to this point if

- The email does not correspond to a registered account (hence the REG_ACCOUNT_NOT_FOUND name), or
- The attempt to resend the confirmation failed for some reason
The error handling logic should probably distinguish between these possibilities, and return different codes, so the client could give more applicable instructions.

entrotech (approved):
I'm going to go ahead and approve, since its already a net improvement, though see my comment below.

nclairesays:
@entrotech noticed it and fixed it in my next PR

Issue #719, opened 2021-01-20
tool tip icon for earned points should be consistent with target points


nclairesays:
When the app is a specific width, the tooltip icon under earned points is not consistent with the one under target points

![image](https://user-images.githubusercontent.com/45305029/105139816-a7706300-5aab-11eb-8ef4-406a4244ebde.png)

jahpe777:
Tooltip icon for earned points is now consistent with target points #719

![Screen Shot 2021-03-08 at 9 40 29 AM](https://user-images.githubusercontent.com/40730303/110359479-64823480-7ff2-11eb-8595-fdd2d3ed74a0.png)

ExperimentsInHonesty:
@jahpe777 can you update the labels for this issue and then close it please.

ExperimentsInHonesty:
No longer an issue because we moved to dial format

PR #772, opened 2021-03-08
719 tooltip earned consistent target

jahpe777:
Tooltip icon for earned points is now consistent with target points #719

Screen Shot 2021-03-08 at 9 40 29 AM


nclairesays (changes requested):
Don't think that resolved the issue.
image

Something you could do: on the div that is containing the "earned/target points" text and the tool tip, put a min-width so that it will never go smaller than a certain size. that way the tool tip and the text will always be on one line.

nclairesays (approved):

PR #817, opened 2021-04-27
762 password requirements

### LINKED ISSUE #762: Enforce Secure Password
### Overview
We should enforce users to create strong passwords.

### Action Items

Make the password so it requires: 
- [x] 12 Characters 
- [x] 1 Special Character
- [x] 1 Capitalization 
- [x] 1 Number

Update wording on red warning box under "password" on  /register page to "

- [x] "Your password must have at least 12 characters, at least one number, one special character, and one capitalization"

jahpe777:
Closes #762 - Implemented password requirements

Screen Shot 2021-04-27 at 9 00 05 AM


nclairesays (changes requested):
Need to add requirements to the ResetPassword component also

nclairesays:
Password validation/Regex doesn't allow all special characters. eg. `asdfghjkl;A` did not work.  I think it's looking for specific ones. Currently, a user would not know which special characters are allowed. Either we need to write in the helper text which characters are allowed and/or allow more characters. both is preferred.

entrotech (changes requested):
I would agree with Claire. If we are going to force them to enter at least one special character, we at least need to specify which characters we consider to be special.

nclairesays (changes requested):
Still doesn't work.. I tried 
```
aab11111-1111D
```
and it gives me error.

I will also try to help with this. Regex with Javascript seem to not behave as expected.

Issue #850, opened 2021-06-10
Add User defined strategy description on Summary Page


KPHowley:
### Overview

We need to add a new section on the final summary page that displays the user's description of the User defined strategy. 

### Action Items

- [ ] Add a line called "User Defined Strategy Description:" to the final summary page
- [ ] Display the description in the column under the other project information or inputs
- [ ] Make so this is formatted correctly for long blocks of text 

### Resources/Instructions
https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=1744%3A3531

Jonathanko52:
-Move User Defined Strategy description under TDM measures selected header instead of its own header.

Shaeeka:
[Figma file](https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=1744%3A3531), please refer to the Summary page Final Draft

-  Updated the color to match the style guide
-  Fixed the spacing of the Project Detail section

![image](https://user-images.githubusercontent.com/88021166/131802606-71e423c3-97db-4eed-b56a-1634fb8e2669.png)

seenaiype:
@Jonathanko52 , as discussed, this ticket is for the userDefined Strategy piece on the summary page.

Shaeeka:
![image](https://user-images.githubusercontent.com/88021166/132611841-0879b82e-1dab-401d-a134-fbb4b5afa72d.png)

Jonathanko52:
1. Move summary text from center align to left align.
2. Change title to "User-Defined Strategy Details."
3. make spacing between 'User-Defined Strategy' and 'User-Defined Strategy Details' consistent with the others.
4. Remove the floating "5" That appears in the User-Defined Strategy row, between the two columns.

seenaiype:
Changes we talked about 

1. Update the header to "User Defined Strategy Details"
2. Remove the extra points that is showing up on the user defined strategy row
3. make the spacing consistent
4. Left allign the user defined strategy details text.

Please, check [Figma file](https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=1744%3A3531), please refer to the Summary page Final Draft, for anything i might have missed.

Shaeeka:
Changed the border's design (corners to 0, stroke to 1, shadow to inside) to match it with the original Earned and Target Point Border
[...]

Issue #873, opened 2021-07-17
TDM Summary Page redesign 


VJ1285:
### Overview
Redesign Page 5 - TDM Summary page

Action Items:
- [x] Create Mock-up (or mock-ups) for Final page
- [x] Update the title to show - Project Name - Address
- [x] Switch the target and earned points, so its consistent with the sidebar
- [x] Move the project level and parking provided baseline % to the top section of project details, maybe as an extra column after Parcel and Version #
- [x] Update _Version #_ to  _Alternative #_
- [x] Results to only show the points section, Color code the numbers to red and green, based on target reached
- [x] Move _Land Use_ section to _Project Specifications_
- [x] Add a title for _Parking details_ for the baseline parking and Parking provided

- [x] Message around target point achieved or not. 
                > If Target points are achieved, update the success toast to state. That they can Print and submit this page for getting their permit.
               > If target is not achieved - Come up with best wording to notify users they have not meet their target points and can return to select more strategies
- [x] Layout to be - Project Name, Project Details , Results, TDM Measures Selected, Project specifications - Land use and Parking 
- [x] Add a button for Save & Print
- [x] Keep formatting and color/style in line with other notifications on page
- [x] Show design or designs to product team for visual review

![image](https://user-images.githubusercontent.com/20568596/128176371-ae971118-1a9c-447d-bae8-34ad932fb206.png)

-------------------
Below ate the issues that need to be addressed on Page 5

Row 28 - Users unclear what to do if they did not succeed, as this is not displayed
Row 29 - Users unsure as to whether the project level is important and why the "parking provided/baseline" was emphasized 
Row 30 - Users are confused about what to do with the information presented on this page
Row 31 - Users are concerned about the approval process (Are they really "good" after going through the calculator)

### Ideas
Row 28 - Need to convey via color coding and text that the user met the target points
Row 29 -
Row 30 -
Row 31 - Need to convey to the users that the information will be reviewed by LADOT prior to approval.

### Resources/Instructions
https://docs.google.com/spreadsheets/d/1SEd8uPOwOmEMdsDR_pylZ2tiByopzPrmlqB5Y4Ka4xc/edit?usp=sharing

KPHowley:
@VJ1285 - I believe all of this had been addressed with other issue  - let me know if anything was not turned into it's own issue.

shinhope:
draft completed
[...]

PR #896, opened 2021-08-06
850 add user strategy to summary

Jonathanko52:
#850 User strategy entered on 4th page now appears on 5th(Summary) page.
Screen Shot 2021-08-05 at 6 01 17 PM

### DIFF
--- a/client/src/components/ProjectWizard/WizardPages/ProjectSummary.js
+++ b/client/src/components/ProjectWizard/WizardPages/ProjectSummary.js
@@ -193,7 +193,7 @@ const useStyles = createUseStyles({
     minWidth: "180px",
     maxWidth: "100%",
     marginRight: "110px",
-    marginLeft: "110px"
+    marginLeft: "80px"
   }
 });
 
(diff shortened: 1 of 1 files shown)


nclairesays (changes requested):
@Jonathanko52 
- The only time a user would write a description in the user defined strategy (going to use UDS) is if they write in their own points for that field. It doesn't make sense to have a specific section for just  UDS.
image

image

Instead, it should be under Project Measures.

image

-------------

Something else I noticed that should have been added to this issue is to include the points for the UDS too.  Please add the points similar to how the "start a new TMO" "join TMO" and then have  the UDS's description below that.

KPHowley:
Hi @Jonathanko52  -  I am going to push this back into progress, and we should make the changes according to Claires comments!

Jonathanko52:
Screen Shot 2021-09-01 at 9 26 12 PM
Border, margins changed. Please let me know if this is closer to what you had in mind.

Shaeeka:
Hey John,
You are close, I have attached the screenshot with the figma link in #850
and #873. The width of the summary section in the figma is smaller than the
webpage. I would prefer the border of the text frame to be shade of grey
E7EBF0 and remove the indentation keeping it parallel to the header. If you
need more explanations please feel free to slack me.
[image: image.png]

Thanks,
Shaeeka

Jonathanko52:
Requested changes made.

Issue #858, opened 2021-07-07
Child Care and School Safety Campaign issues


awikstrom-ladot:
### Overview

The Child Care strategy is greyed out on all uses, even though it is applicable to all uses except housing. 
School Safety Campaign is select-able under multiple uses, and should only be select-able with schools

### Action Items

- [x] Allow Child Care strategy with all projects, except for housing
- [x]  Exclude School Safety Campaign strategy from Employment, Retail, and Housing uses.

nclairesays:
@awikstrom-ladot your original message says 
> School Safety Campaign ...should only be select-able with schools

The action item says: 
> Exclude School Safety Campaign strategy from Employment, Retail, and Housing uses.

Does this mean that the other types like, Hotel, warehouse, medical, other, and major  can select School Safety Campaign? Or is it literally only checkboxes under "Schools" that can choose that measure?

awikstrom-ladot:
This means that only schools can select School Safety Campaign as a strategy. Unless school is selected when an applicant is inputting their project, this measure should be greyed out

PR #897, opened 2021-08-06
updated child care display rules and school safety campaign strategy #858

jahpe777:
- Allowed Child Care strategy with all projects, except for housing
- Excluded School Safety Campaign strategy from Employment, Retail, and Housing uses 
#858


nclairesays:
I'm going to go ahead and migrate the first migration file for child care. However, I will wait on an answer for schools as the original github issue is not clear. 

https://github.com/hackforla/tdm-calculator/issues/858#issuecomment-895783708

nclairesays (changes requested):
Only schools can select school measures, per comment by ladot on the original issue

(Edit the the same file with the update, you can run `npm run flyway:clean` on your LOCAL db to get rid of the previous changes as we don't want to include them, then `npm run flyway:migrate`to migrate the new changes)

PR #1087, opened 2022-03-10
Move terms and conditions modal to all pages

### LINKED ISSUE #1082: Move terms and conditions pop up
### Overview
The terms and condition pop up comes up when a user goes to the wizard (calculation/1) for the first time this should pop up regardless of what page the user visits first.  If there is another pop up on this page this modal should be focused on first and the rest of the paged should be greyed out. 

### Action Items

- [x] Change logic so the terms and conditions pop up comes whenever a user visits any page on the site. 
- [x] keep logic so the pop up is no displayed to repeat visitors.

fyliu:
Terms and Conditions modal now shows up on any page. Once dismissed, it will not show again for any page.

Closes #1082

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -11,6 +11,7 @@ import Header from "./components/Header";
 import Footer from "./components/Footer";
 import About from "./components/About";
 import TermsAndConditionsPage from "./components/TermsAndConditions/TermsAndConditionsPage";
+import TermsAndConditionsModal from "./components/TermsAndConditions/TermsAndConditionsModal";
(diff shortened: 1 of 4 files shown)


fyliu:
Screenshots of the T&C modal on several pages

About Page
2022-03-10 13 43 14 localhost d8a4b20a49ad

Public Comment Page
2022-03-10 13 43 00 localhost 14b9db4191fd

Create Project Page
2022-03-10 13 42 40 localhost eb43efd083a0

Login Page
2022-03-10 13 42 24 localhost 3b49f7ec2d26

Issue #1024, opened 2021-12-09
Update final summary page with dropdown design to show mulitple AIN's


KPHowley:
### Overview
We have a new system allowing for multiple AINS and have updated the design for the final summary page to show this. 

### Action Items

- [ ] Add drop down box on side of AIN display
- [ ] Allow for three AINs to be displayed in a line and then hidden afterwards
- [x] Normal state will only display one line
- [x] AINs displayed in same order as from the Create Projects page

### Resources/Instructions

![Frame 9](https://user-images.githubusercontent.com/82553907/143803684-c546a3fa-40cc-49f9-a9c8-74bc09553403.png)

Mockup for AIN/APN summary page. The section will display three AIN/APN numbers at a time by default before it expands.

See mockup here: https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=5517%3A19644
See prototype here: https://www.figma.com/proto/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=5517%3A19591&scaling=min-zoom&page-id=4189%3A15774&starting-point-node-id=5517%3A19580&show-proto-sidebar=1
_Originally posted by @cduong5 in https://github.com/hackforla/tdm-calculator/issues/988#issuecomment-981257060_

cduong5:
@KPHowley Looks good! Don't have additional action items to add.

ExperimentsInHonesty:
@fyliu Please provide update
1. Progress: "What is the current status of your project? What have you completed and what is left to do?"
2. Blockers: "Difficulties or errors encountered."
3. Availability: "How much time will you have this week to work on this issue?"
4. ETA: "When do you expect this issue to be completed?"
5. Pictures (if necessary): "Add any pictures that will help illustrate what you are working on."

fyliu:
There's no progress yet. I just claimed this issue. ETA is next week but after the next team meeting 3/16.

fyliu:
Clarification needed: Do we want to have 3 AINs per line or is that the max AINs per line?
See this screenshot where 3 AINs don't fit on a single line. This is before adding the chevron icon on the right
![2022-03-15 22 51 18 localhost 6cf8c926125d](https://user-images.githubusercontent.com/1160105/158525335-b837aa76-e05a-4f15-8d3f-df1653735352.png)

fyliu:
1. Progress - 
2. Blockers - 
3. Availability - 
4. ETA -

[...]

PR #1105, opened 2022-03-26
Working 1024 issue

poorvarathore05:
summaryPage

summaryPage2

Remove the DropDown and Update final summary page AIN's no. in a row.

### DIFF
--- a/client/src/components/ProjectWizard/WizardPages/ProjectSummary/ProjectInfoList.jsx
+++ b/client/src/components/ProjectWizard/WizardPages/ProjectSummary/ProjectInfoList.jsx
@@ -6,9 +6,7 @@ const useStyles = createUseStyles({
   projectInfoDetailsSubContainer: {
     display: "flex",
     alignItems: "baseline",
-    width: "50%",
-    paddingBottom: "1.1em",
-    maxHeight: "50px"
+    width: "50%"
   },
   projectInfoCategory: {
     fontFamily: "Oswald",
@@ -26,7 +24,8 @@ const useStyles = createUseStyles({
   AINValuesContainer: {
     display: "flex",
     flexWrap: "wrap",
-    flexDirection: "row"
+    flexDirection: "row",
+    gap: ".3em"
   },
   AINValues: {
     minWidth: "100px",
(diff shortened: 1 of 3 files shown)


client/src/components/ProjectWizard/WizardPages/ProjectSummary/ProjectInfoList.jsx
```diff
@@ -0,0 +1,65 @@
...
+  projectInfoDetails: {
+    color: "#0F2940",
+    fontFamily: "Calibri Bold",
+    fontSize: "16px"
+  },
+  AINValuesContainer: {
+    display: "flex",
+    flexWrap: "wrap",
+    flexDirection: "row"
+  },
```
fyliu, line 29: Looks good! Maybe consider doing justifyContent: "space-between" in the AINValuesContainer rather than minWidth in AINValues. Also maybe add gap: "5px" or some value like that so space out the AIN values.

fyliu (changes requested):
Overall the code looks good.
The only thing I found was that having more than 2 rows of AIN values would push the bottom row of the section out of its box. Some parent container probably has a set height that should really be expandable instead.
2022-03-25 20 30 16 localhost f4b9b821021d

client/src/components/ProjectWizard/WizardPages/ProjectSummary/ProjectInfoList.jsx
poorvarathore05: Thanks for highlighting this thing. I missed that part. Certainly, I will work on that and will fix it.

poorvarathore05 (commented):
Hello @fyliu 

I made some changes, Basically, maximizing the width for the parent component.
And as you suggested adding the justify-content, so that is not showing values properly when there are only 2-3 AIN values. So I removed that

fyliu (commented):
Thanks for the fixes. I made a couple suggestions.

poorvarathore05:
Hello @fyliu,

Made the changes as you highligted.

fyliu (approved):
Looks good. Thanks for working on this issue!

fyliu:
Closing #1024

PR #1178, opened 2022-06-03
fix #1022

### LINKED ISSUE #1022: Fix Horizontal alignment of columns on Target Points Page
### Overview
Decide on reasonable horizontal alignment of visual elements on the Target Points Page of the wwizard and make changes to implement, including labels, numers and units (where applicable

### Action Items
- [ ] Decide on how each element should be aligned
- [ ] Implement in code.

### Resources/Instructions
image

poorvarathore05:
Capture2

### DIFF
--- a/client/src/components/ProjectWizard/RuleCalculation/RuleCalculation.js
+++ b/client/src/components/ProjectWizard/RuleCalculation/RuleCalculation.js
@@ -10,7 +10,7 @@ const useStyles = createUseStyles({
     margin: "1em",
     display: "flex",
     flexDirection: "row",
-    justifyContent: "center"
+    justifyContent: "flex-end"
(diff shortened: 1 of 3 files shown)


entrotech (changes requested):
In virtually every application that I know of (including ours), columns of numbers are right-aligned or are horizontally aligned on the real or implied decimal point (if applicable), for example:
Screenshot 2022-06-04 124457
The summary page also shows Parking Provided / Baseline, but rounds it to the nearest integer, avoiding the question of how to horizontally align decimal numbers.
The Parking Provided / Baseline number on the Target Points page can have up to two digits after the decimal, and I think aligning to the right-most digit (regardless of the decimal) is probably best.

PR #1183, opened 2022-06-04
Changed time delay from 500ms to 750ms

entrotech:


entrotech:
These changes are on different PR.

PR #1272, opened 2022-11-17
Added DB migration for the faqCategory table, server side files

jahpe777:
Added DB migration for the faqCategory table and started creating server side files in order to retrieve faqCategory on the client side.


entrotech (approved):
I haven't looked much at the front-end code, but will approve to get the back-end and database working.

Issue #1261, opened 2022-11-03
Update the User-Facing FAQ Page Implementation to include categories


entrotech:
### Overview
Update the User-Facing FAQ Page Design to include categories

### Action Items
- [x] Decide on, and implement the required database changes. Normally, I would create a separate table FaqCategory and add a foreign key from the Faq table to the FaqCategory table, However, with the way the administration screen for FAQs is intended to work, it might be easier to represent categories as rows in the Faq table with a type of "Header" (vs "Question" for the normal Faqs). Should discuss the options with @entrotech before implementing.
- [x] Modify the screen functionality to incorporate the categories as headers
- [x] Try to match the wireframes provided by Renee under the Resources section.

### Resources/Instructions
Mockup-User Facing FAQ w/categories
https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=7089%3A24872

Reneedesigns:
Screenshots: Selected screens per 11/2 team meeting
Categories within banner with questions collapsed and some answers expanded

Figma link to screens
https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=7089%3A24872


https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=7089%3A24872

https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=7089%3A24872

@entrotech @jahpe777  @Biuwa @syria

jahpe777:
Updated the User-Facing FAQ Page Implementation to include categories #1261

PR #1301, opened 2023-01-12
1261 update user facing faq categories

jahpe777:
PR for review of Faq expand/collapse functionality


jahpe777:
Closes  #1261

Updated the User-Facing FAQ Page Implementation to include categories

entrotech (commented):
@jahpe777 I did some simplification and re-arrangement of the code. We only want to be fetching the FAQs and categories once, since re-fetching takes time, causes some annoying flickering of the UI, and, once we get to the editing part would undo any pending changes before they are saved.  So, I changed this back to populating the faqCategoryList from the database only once and using it as the "source of truth" for the whole FAQ page.  The expandFaq, collapseFaq, expandAll and collapseAll functions were re-written to dig through faqCategoryList to flip the state of the expand properties on the correponding faq objects inside the faqCollection.faqs arrays.  We then don't need to have an faqList object in the FaqView state.  The trick is getting changes to the expand properties to re-render the appropriate objects, which I did by adding key props on the list of categories in FaqCategoryList and FaqCategory components.  It all seems to be working pretty well now.

jahpe777:
@entrotech that is a good point, looking ahead to the editing part. 

My initial approach regarding all of the expand/collapse functions was to dig down into the faqCategoryList but I could not figure out how to reach the corresponding index within faqCategoryList.faqs. It seems you were able to do that by adding key props on the list of categories in FaqCategoryList and FaqCategory components.

My second approach (the one I went with) was to simply reconsolidate the faqCategoriesList with the faqList after clicking the expand/collapse functions but as you pointed out, this would be problematic once we reach the editing part and did take longer to re-fetch. 

Another thing I added was a migration to simply add the "Q:" to each question per @Reneedesigns ...designs. https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=7089%3A24872
I don't see this reflected yet on the dev site.

PR #1337, opened 2023-03-09
Non-typography changes made

Jonathanko52:
"Typography" issues are turning out problematic. Font families defined in the Figma don't seem to match up with the CSS. i.e. Headers in the figma appear to be have changed to "Calibri" in the new figma model, but actually entering that value appears to produce drastically different results 

Image from TDM 
Screen Shot 2023-03-08 at 9 14 05 PM

From Figma:
Screen Shot 2023-03-08 at 9 12 08 PM

Are the font weight and family accurate?


Jonathanko52:
Changes to the Typography as specified in the figma have been implemented (Barring "Letter." I'm not sure what that property is). Actual results seem to vary depending on browser, and I can't confirm it with my current setup.

I recommend that this pull request by merged, as all the other changes have been completed. The typography issue can be revisited later.

PR #1355, opened 2023-04-06
1348 Theme Integration - Create Project page 1

shanesween:
Added theme integration for page 1 - create project

In addition, this PR updates all red text to new warning color


entrotech (approved):
I noticed at least two differences in appearance from the current production version:

- The disabled nav button has a grey background in prod, but a translucent green in the new version.
- The Wizard Page titles are Calibri Bold font family in prod - the new version may be more correct as Calibri Font Family, 28 pt, Font Weight 700.

This PR is definitely a move in the right direction, so I'm just merging.  We should probably get the UI team to review the changes to make sure they are happy with it.

Issue #1305, opened 2023-01-14
Implement the Admin Version of the FAQs page


entrotech:
### Dependency 
- #76 
-  #1261

### Overview
- Implementation of design issue #76. This allows an admin to enter and update FAQ questions and categories. 

### Action Items
- [ ] May need to define detailed steps to this project - it is fairly challenging, with drag-n-drop in a hierarchical data structure.
- [ ] Remove dependency on #1304

mojimoh:
@entrotech
Please provide update
1. Progress: "What is the current status of your project? What have you completed and what is left to do?"
2. Blockers: "Difficulties or errors encountered."
3. Availability: "How much time will you have this week to work on this issue?"
4. ETA: "When do you expect this issue to be completed?"
5. Pictures (if necessary): "Add any pictures that will help illustrate what you are working on."

entrotech:
Implementing the Answer editing: For this, some of the existing answers have HTML tags in them, like links,  As long a s we do that, we may as well implement other formatting elements like ordered and unordered lists, bold text, paragraphs, etc.  to do this, I had to modify the design to give the user an "HTML Editor" to use when they choose to edit an answer.  This is mostly working at this point.  However, when you are in edit mode and click on a Category Name, Question or Answer to begin editing, it should automatically give focus to the field you want to edit,  but I have not implemented that yet, and choose to defer fixing that unless/until the team agrees that the overall concept has a good user experience.

Implementing drag0n-drop has been very challenging. There are several libraries available to implement drag-n-drop, and I spent about two days of work trying to use [dnd-kit](https://dndkit.com/), only to find that once I got it working, the draggable parts (i.e. the panels for editing FAQs or Categories could not respond to mouse clicks of keyboard actions. So I had to search for another library to use, and settled on [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd), which  took about 2 days of work to get working. This was pretty challenging because there was an [egghead course](https://egghead.io/lessons/react-course-introduction-beautiful-and-accessible-drag-and-drop-with-react-beautiful-dnd) on how to use it, but their examples were written with class-based react components (vs functional components we use), which made the handling of react refs quite different than what we need.  I did find a poorly-written, but still [helpful article](https://dev.to/kyleortiz/getting-started-with-react-beautiful-dnd-using-functional-components-50d0) that helped me adapt the egghead example to functional components.

At this point, both the editing and drag-n-drop are working pretty well, so the remaining parts are:

[...]

PR #1397, opened 2023-06-28
1305 faq admin refactoring

shanesween:
This Pull Request includes the following modifications:

- Refactoring local state of the faqCategories & faqs
- Ability to add/edit/move/delete categories
- Ability to add/edit/move/delete faqs
- Ability to save form to the DB
- Moved all logic to a single hook component (to de-clutter `<FaqView />`)
- Migration files that add a new sql procedure `insertAll`. This new procedure will delete the data stored in `faqCategory` TABLE and replace it with the `form state` onSave. With this work, I have added `faqs` as a `column` in `faqCategory` TABLE. This eliminates the need to store `faqCategories` and `faqs` in separate tables (and eliminate the need to make requests for each and augment the FAQs into the categories state).


client/package.json
```diff
@@ -29,6 +29,9 @@
     ]
   },
   "dependencies": {
+    "@dnd-kit/core": "^6.0.8",
```
entrotech, line 32: The three dnd-kit packages are remnants of my first attempt at dnd, and are not needed. The react-beautiful-dnd package is all you need.

entrotech (changes requested):
I downloaded and ran your branch. It is working pretty well. Here are some things that need work:
- You are able to edit categories and FAQs on-screen and save the changes when you aren't even logged in, if you make changes to categories and/or FAQs and press the "Edit Section" button twice.
- When I ran the DB migration on my local database, the migration ending in "1230" failed on the first command with a SQL error that a Foreign key referenced the FaqCategory table, so it could not be deleted.  I was able to fix it by swapping the first two SQL commands in that script - deleting the FaqCategory table before the Faq table, then running `npm run flyway:repair` to accept the modified migration script and then `npm run flyway:migrate` to finish the migration.
- The database migration completely wipes out the existing Categories and FAQs, which took a lot of effort to create.  We do have a migration script that could re-create the Faq and FaqCategory table data, except that you re-organized the table structure, so we need to map the FAQ table data into the FaqCategory.faqs JSON column, or create a different DM Migration script to re-populate the table with the initial FAQs.
-  The "Edit Section" (not a good choice of wording - what is "Section" supposed to mean in this context?) should only be visible if you are an admin.
-  Unless you are an admin and have pressed the "Edit Section" button, the Category, Question and Answer fields should all be read-only.
- The read-only rendering of the answers seems to render regular text the same as bold text - this seems to be because of the styling of the read-only answer - we probably need to remove the bold style applied to the read-only answer to fix this?
- The Expand All / Collapse All feature design has problems inherent in the design given to you.  If only a few FAQs are expanded, you should have both the Expand All and Collapse All options. The way it is implemented now, you only get one of these options, and it (at least sometimes) does the opposite of what the label says.

shanesween:
Additional changes have been pushed to this PR that resolves @entrotech 's comments (#'s 1, 4, 5, & 7). I need to discuss how we want to handle the DB migration changes (#'s 2 & 3) going forward - will discuss with @entrotech on 7/26 meeting. 

This leaves #6 still open. I am working through this issue. Currently, the FAQ's answer is being saved as a string and rendered as a string; so the HTML modifications from the enriched text (such as bolded font, or hyperlinks, or other enrichment options) are not being rendered correctly. I will continue this work - hoping to have this issue #1305 fully completed by next week, at the latest the following week. 

cc: @Biuwa

PR #1401, opened 2023-07-06
1282 create accessible modal

### LINKED ISSUE #1282: Create Accessible and Reusable Modal Dialog Component and Refactor Existing Dialogs to Use It
### Overview
We need to use a common dialog component for implementing modal dialogs in the application in order to reduce duplicate code and ensure consistency in the look and feel of the various dialogs. Currently, there are several modal dialogs in the application, each of them implemented independently, with differing behavior and styles.

### Action Items
- [ ] Create the common dialog component.  The behavior should be:
  - Auto-focus on the first focusable body element when the modal is opened.
  - Trap focus on the modal so that it doesn't leave the modal elements until the modal is closed.
  - The Escape key should close the modal without taking any action.
  - The Enter key should have standard behavior (i.e., if a button has focus, clicking the button)
  - Standardize the appearance of the Header's close button.
- [ ] Use it to replace the Project Delete dialog
- [ ] Replace the "Leave the Project Wizard" Dialog, 
- [ ] Replace the Delete FAQ dialog, 

### Resources/Instructions
See https://medium.com/@seif_ghezala/how-to-create-an-accessible-react-modal-5b87e6a27503 for a good implementation.

Examples:
Checklist Modal Dialog:
image

Leave the project Wizard Dialog:
image

Delete Project Dialog:
image

Terms and Conditions Dialog
image

Jonathanko52:
Fixes #1282


entrotech:
I found the combination of the ApplicationModal and ModalData components to be very cumbersome and inflexible. If you wanted to add a Modal with a new type of content that didn't fit within the structure that ApplicationModal supports, you would need to modify ApplicationModal, ModalData and the component hosting the ApplicationModal.

I re-wrote this with a standard modal dialog that only provides the backdrop, modal content area, close box and associated behaviors like keystroke handling closing when the backdrop is clicked. To create a specific modal instance, you then use the ModalDialog and specify the children props applicable for your specific use, which is completely tailored as you need. I created PR #1422 with this implementation and replaced all the dialogs. The only part that doesn't really work well is auto-focus when the dialog is first opened. It turns out that only the CopyProjectModal has any controls except buttons on it, and it seems hard to auto-focus on a button (maybe there is a reason for this, I don't know).  Maybe the focus-trap-react library handles this - I will need to experiment with re-introducing it, and think more about how the initial focus should work.

Anyway, I'm closing this PR without merging, in favor of PR #1422. There will certainly be follow-up issues to resolve deviations from the original wireframes for the various dialogs, and perhaps detailed behavior of the standard dialog.

Issue #1395, opened 2023-06-26
Create a "Delete User" API Endpoint for Security Admin Only


azajzon:
### Overview

There is currently no way for the security admin to control the deletion of users. A delete API endpoint must be created to meet this requirement, but it should work as a "soft delete" in order to be able to reverse the action if necessary for a set time period.

#### Frontend
The security admin should be able to log in and see his users with a delete option for each user. 

#### Backend
Create archiving and deleting api endpoints and SQL migration for affected tables.

***

#### Backend Action Items
- [x] Create a api endpoints for deleting users and accounts (soft delete and hard delete)

#### Frontend Action Items
- [x] implement UI to interact with backend api

agosmou:
Adding John to this issue for visibility and discussion 

@entrotech

agosmou:
Possible Implementations assuming 30 day expiration after soft delete
1. a user is archived for 30 days by simply not displaying on the FrontEnd
    a. This would require editing the other api endpoints such as the GET all users in order to hide the "deleted" users
2.  an archive table is created in the database, so when a user is indicated to be deleted, he will be moved to the archive table for 30 days, where he will be deleted after the time limit expires.
    a. This would isolate this endpoint and the deleted users to prevent bugs/errors

agosmou:
Regarding the reversal action:

We should discuss how this would look because it could create more scope for the front end, such as creating an archive page that shows all deleted users and the timeline they have until irreversible deletion with a button to reverse the action.

The reversal itself could be implemented as:
- a function with an event listener "onclick" that calls a "reversal" api endpoint

agosmou:
WIP

https://github.com/hackforla/tdm-calculator/commit/8cb57229292c43418dc4b50da5bae2cd269940e3

agosmou:
If we implement a soft delete by adding a column with a date and time such as "deleted_on"

then all devs must always remember to filter out the rows from the table where "deleted_on" is not null....

agosmou:
Looking into this further, the stored procedures can create another layer of complexity when implementing this db change. 


Perhaps an alternative to the soft delete could be a hard delete with 2FA and have the user type in the name and email of the user to confirm their deletion.

agosmou:
Met with team, need info from product to know how long users and their data should persist. (e.g. 30 days, 1 year, indefinite, etc)

agosmou:
[...]

PR #1449, opened 2023-09-11
Adds end-to-end feature for archiving and deleting user accounts

agosmou:
### **Summary** 
This PR solves #1395 
Implemented an end-to-end feature, creating a workflow that allows security admins to archive, un-archive/restore, and delete users along with their corresponding projects. 

### **Features** 
- **Archiving Users:** Security admins can now archive users. 
- **Un-archiving/Restoring Users:** Security admins can also un-archive and restore users.
- **Deleting Users and Projects:** Security admins can permanently delete users and the user's respective projects.
- **UI Enhancements:** Added buttons and confirmation modals to ensure admins understand the implications of their actions.

### **Technical Details** 
- Created a migration script to alter the tables "Login" and "Project" to add an `archivedAt` column that is set to `NULL` by default or to a timestamp if queried
- Created new, protected API endpoints: 
    - Accounts: 
        - PUT `/api/accounts/:id/archiveaccount`
        - PUT `/api/accounts/:id/unarchiveaccount`
        - GET `/api/accounts/archivedaccounts`
        - DELETE `/api/accounts/:id/deleteaccount`
    - Projects:
        -   GET `/api/projects/archivedprojects`
- Added new components in the frontend to handle the workflow.
- added new frontend routes to show all archived users or all archived projects

### **How to Test** 
1. Login as a security admin.
2. Navigate to the `/Roles` page.
3. Next to each user, there should be an 'options' button(vertical ellipsis).
4. Clicking the 'options' button will render a PopUp that can be clicked to archive the user. This will populate the `archivedAt` field for the user and their projects; thus, 'archiving' the user. Security Admins cannot be archived or deleted. The user is also not able to self-archive or self-delete.
5.  To see the archived users, navigate to the `/archivedaccounts` by following the link. Here, the security admin can 'un-archive and restore' a user OR permanently delete by clicking on the arrow or trash icons and their respective PopUp's. These actions will take effect on the user and their corresponding projects.
6. To see all archived projects, navigate to the `/archivedprojects` by following the link. This page only shows all archived projects - no actions. To return to the `/Roles` page, follow the link on the top of the page. 

### **Feature Notes**
- Deleting a user is irreversible - there is a currently a multi-step-process to delete a user to prevent accidental removal.
    1. select 'options'
    2. use popup to confirm user is to be archived
    3. Visit the archive page
    4. select 'delete'
    5. use popup to confirm user is to be permanently deleted
- Only Security Admins can archive, un-archive/restore, or delete users.
- The logged-in user cannot self-archive or self-delete
- Archived users cannot login

### **Future Improvements** 
- We could consider adding 2-Factor Auth such as email confirmation or PIN to delete users

#### 

firefox_uOiRVYKpGB

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-# lerna run lint
+lerna run lint
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-lerna run lint
+# lerna run lint
--- a/client/.prettierrc.js
+++ b/client/.prettierrc.js
@@ -5,5 +5,5 @@
 module.exports = {
   trailingComma: "none", // deafult changed in prettier 2+
   arrowParens: "avoid", // default changed in prettier 2+
-  endOfLine: "lf"
+  endOfLine: "auto"
 };
--- a/client/.prettierrc.js
+++ b/client/.prettierrc.js
@@ -5,4 +5,5 @@
 module.exports = {
   trailingComma: "none", // deafult changed in prettier 2+
   arrowParens: "avoid", // default changed in prettier 2+
+  endOfLine: "lf"
 };
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -22,6 +22,8 @@ import Login from "./components/Authorization/Login";
 import Unauthorized from "./components/Authorization/Unauthorized";
 import Admin from "./components/Admin";
 import Roles from "./components/Roles";
+import ProjectsArchive from "./components/ArchiveDelete/ProjectsArchive";
+import RolesArchive from "./components/ArchiveDelete/RolesArchive";
 import FaqView from "./components/Faq/FaqView";
 import ResetPassword from "./components/Authorization/ResetPassword";
 import ForgotPassword from "./components/Authorization/ForgotPassword";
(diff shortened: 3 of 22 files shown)


agosmou:
### **Branch Notes**
- lerna was giving me issues when commiting, so I had to comment it out
- I developed using a local db so the migration must be run on the shared development db
- 'prettier' was giving me tons of issues as I developed on my windows machine, especially with line-endings so I tried to mitigate as well
- The migration I did, using my local db, caused a large change in the `/server/report.html` and `/server/report.json` files that we may want to remove before merging this branch

agosmou:
Ready for review!

Do note the comment above:


> ### **Branch Notes**
> 
>     * I developed using a local db so the migration must be run on the shared development db
> 
>     * 'prettier' was giving me tons of issues with line-endings as I developed on my windows machine so I made a fix on the /server and /client directories to mark the endOfLine property to `auto`
> 
>     * The migration I did, using my local db, caused a large change in the `/server/report.html` and `/server/report.json` files (~6,000 lines)

entrotech (approved):
This PR is FANTASTIC! I ran the migration against the shared dev db and it worked great, then tested all the features you implemented and they all worked perfectly, as far as I can tell.

There is a bit of overlap of your project archive feature with a different Issue that allows users to move their own projects to the trash and eventually delete them, but the purpose is a bit different, so we need to figure out if the admin delete/archive and user delete/archive for projects can just remain separate or not.

I had seen the report.html and report.json files crop up before, but didn't look into where they came from. It turns out the Flyway, the DB migration tool auto-generates these files.  I read enough to know that they document the migration changes and you can somehow run reports on the migrations using them as source data, but we might look into their utility and see if we want to use them and figure out how to disable their generation if we don't.  If we do want to use them, we should probably change the flyway config to give them better names, and tuck them away in the /server/db folder so they are out of  the way.

Karandeep was also having trouble with line endings in Windows. We can talk about the relevant settings tonight.

I'm sure the UI designers will want to make some tweaks to the UI appearance, but the functionality is all there.

Issue #1403, opened 2023-07-07
Tooltip Hover Activation


Noushie:
### Overview

**Note*: Revert back to #1402 for actual hover state changes** 

Goal is to increase visibility of the 'information' icon, the hope is to have the hover state activated as the user places their cursor in any of the following area (denoted by the highlighted bar below):

![image](https://github.com/hackforla/tdm-calculator/assets/115515867/13edcfa7-1baf-44d1-8373-b589c9369ac3)

### Action Items

- Increase distance from term that activates hover state 

### Resources/Instructions

View in Figma [here](https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?type=design&node-id=9846%3A27098&mode=design&t=prphMoNVohaH2WBs-1)

Jonathanko52:
Hi team, I just wanted to check on something"

Not included in the "Action Items" of this issue are new behaviors demonstrated in the figma, such as:
1. graying out the background of the term item
2. graying out of the new "i" icon added in term 1402.

Is this correct?

Noushie:
Hi Jonathan,

Is the following image what you're referring to? 
![image](https://github.com/hackforla/tdm-calculator/assets/115515867/c05352d7-5286-4342-be98-c00b25fc8462)

If that's the case - the 'highlight' is only meant as visual indicator for where we hope to a user would activate the hover state (in other words, the background/icon should not be greyed out). It is only meant to visual depict where we decided the hover state should start based on user feedback. 

Please let me know if you have any other questions/feedback. Thanks!

Biuwa:
Pull Request has been merged and closed per JK and John's review

Issue #1402, opened 2023-07-07
Change Tooltip Hover/Clicked State 


Noushie:
### Overview

**Replace the bolded, underlined hover over terms with an 'i' icon (on pages 1,2,4):**

![image](https://github.com/hackforla/tdm-calculator/assets/115515867/8557d697-2007-4ec0-96a0-1c91ca50a656)

**Once clicked the terms are clicked and the tooltip drops down, it will become bolded:**

![image](https://github.com/hackforla/tdm-calculator/assets/115515867/04e53b54-c4b6-49fc-8acc-74b0e0e0061c)

### Action Items (colours + icons can be found in 'Full Design System')

**Hover state:**

- [ ] Remove bold, underlined hover state 

- [ ] Add a navy blue (#002E6D) 'i' or information icon next to regular text term

**Clicked state:**

- [ ] Change term text colour to 'navy blue' (#002E6D)

- [ ] Bold term text

- 'i' icon remains on screen

- No changes to the tooltip itself

### Resources/Instructions

View in figma [here](https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?type=design&node-id=9791%3A25976&mode=design&t=prphMoNVohaH2WBs-1)

Note*: See #1403 for the state's activation/trigger

Jonathanko52:
Progress: Tracked down the component that needs to be changed and the icon. Component is being re-used across the project wizard now, so I'm trying to figure out how the data is coming in and how to append the icon without adding it to literally every field.
Blockers: None.
Availability: Weekdays.
ETA: One week.
Pictures (if necessary):

Jonathanko52:
Progress: New Hover and onclick behaviro added to first page. Working on second page items
Blockers: None.
Availability: Weekdays.
ETA: One week.
Pictures (if necessary):
![Screen Shot 2023-08-01 at 10 40 27 PM](https://github.com/hackforla/tdm-calculator/assets/30099154/f4e4cb42-8dec-41bf-9698-ae64c5bd37cc)

Jonathanko52:
Progress: New Hover and onclick behavior added to second page. Icons removed for items without descriptions.
Blockers: None.
Availability: Weekdays.
ETA: Completed

Jonathanko52:
Progress: Fixed issues with text bolding. Ready for review.
Blockers: None.
Availability: Weekdays.
ETA: Completed

Biuwa:
[...]

PR #1451, opened 2023-09-13
1402 change tooltip hovered state

Jonathanko52:
### Summary

Adds behavior requests in #1402 and #1403.

###Features

-Removal of old on click behaviors for rules.
-Icon added to rules with additional descriptions. Becomes visible on hover.
-Clicking on a rule will toggle it, changing the text, icon, and tooltip.
-Hover/clickable area size increased.


entrotech (approved):
Name of Commit should be "Remove redundant code from ToolTipLabel.js

There should almost never be conditional logic that depends on which particular rule (as identified by the code property) it applies to.  In this case, there was a gigantic IF/ELSE statement where the two branches didn't even render anything differently as far as I could tell.

entrotech:
@Jonathanko52 Can you review the behavior in the dev environment and make sure that all the desired behavior is still implemented after I removed what appeared to be redundant code?  Thanks!

Issue #1393, opened 2023-06-18
Create a print button with each corresponding project row within the "Projects" page


jahpe777:
### Overview
We need to implement the button and PDF output for printing/downloading the TDM Calculation Summary within the "Projects" page

### Action Items
- [ ] Create a print button with each corresponding project row within the "Projects" page that prints out the summary of each project

### Resources/Instructions
Please reference #[1314](https://github.com/hackforla/tdm-calculator/issues/1314)

jeffmchang:
### UPDATE
<img width="664" alt="image" src="https://github.com/hackforla/tdm-calculator/assets/109321774/82a0382e-f91f-4854-9d44-240e30701ad6">


## 7/29/23:

### *Recap:*

Identified where we needed to make the change in the file, and noticed that the code was practically there already. The `printPDF` component was loaded in and also had the `ReactToPrint` typed in. Started messing around with the `componentRef.current` on `line 399` in `ProjectsPage.js` that controlled the content. Essentially what’s happening is that the two projects are printing the same PDF because the print button updates ALL print buttons to the `componentRef.current` PDF. 

Second bug with this feature is that when you refresh the page, it breaks it completely because once the referenced component is unmounted and re-mounted, the componentRef variable loses track of where the reference is attached to and returns null. This is only an assumption, but I will be looking more into it.

### *Blocker:*

Need to figure out a way to isolate the projects so the print function doesn’t update ALL the print buttons and keeps them with the respective projects. 

### *Next:*

Going to try to access the `currentProjects` ID and attach them to the `printPDF` function so they’re all individually attached to each mapped item.

jeffmchang:
## 7/31/23:

### *Recap:*

Currently dealing with the same issue, can’t seem to separate the `componentRef.current` object from each individual project. Trying to see if I can work around the existing code before trying a different way. Was able to map out the projects so the pdfs are connecting to their own respective project row, but the content is still duplicating over. 

Mounting and unmounting are still posing an issue, will probably need to implement the `useEffect` hook. Not entirely sure yet, still focused on the pdf mapping.

### *Blocker:*

***existing:*** Need to figure out a way to isolate the projects so the print function doesn’t update ALL the print buttons and keeps them with the respective projects. 

### *Next:*

Will be trying to see if I can use a `createRef()` built-in function that will create refs for each and every mapped-out project. Need to make sure the refs identify the correct information.

[...]

PR #1508, opened 2023-11-13
Zeke download csv and print pdf

### LINKED ISSUE #1420: Implement a "Download to CSV" feature
### Overview
We want users to have the ability to download a CSV file for one or more calculations - one row per calculation. To start, this will be accessible by pressing a button on the last page of the wizard, which will produce a CSV with column headings and one row of data. IMPORTANT: The code to implement this feature must not reside in the Wizard Components, but should be implemented as a function that can be called from anywhere in the application with a single method that is passed an array of one or more calculationId values, and downloads a single CSV file with one row per calculation in the array.

### Action Items
- [ ] Add a "Download" button to the TDM Calculation Summary page of the wizard. The design should look like the  existing "Print Summary" button, with a download icon.
- [ ] When pressed, the button should create a CSV file. The CSV format can have a heading of "TDM Calculation Project Summary" on the first row, a second row with the caption "Date Printed" and the current date/time. The third row will be blank, The fourth row will have "column headings", and the subsequent row will have the data for the current calculation. For downloading multiple projects, the subsequent rows will have one row per additional project (though that will only come into play when the feature is added to the My Project page).
- [x] For scalar data (strings, numbers, etc.), there will be one column with a column heading that corresponds to the label from the PDF print format. So the first few will be "Project Name", "Address", "Parcel #", "Building Permit #", and so on. The easiest way to get a complete list of the columns for the Project Name and Project Details sections is probably to create a sample project where every possible input field is filled in with data. For strategies, some of the strategies are multiple choice. In these cases, a separate column will exist for each possible choice. For example, there will be two columns for the Encouragement Program strategy, one with the column heading "Encouragement Program - Education, Marketing and Outreach" and a second with the heading "Encouragement Program - Voluntary Behavior Change Program". The selected option will have a "Y" in the project's row. 
- [x] I created a project using the ladot@dispostable login called "Kitchen Sink Project" that has almost all possible fields selected (except "Mandatory Trip Reducction" and "Cash-out" strategies, since they are mutually exclusive with other strategies). If you look and PDF print-out from this project, it shows most of the possible inputs and results on the Summary.
- [x] You can use the service on the client project.service.js called getById to get all the project data you need. For multiple projects, you would have to loop over all the ids in the array, but we could eventually create another endpoint to get all the projects at once.
- [x] The CSV file format needs to be documented and the document delivered to the CLATS team (Mony Patel, Michelle Leung and Jason Kwok).

### Resources/Instructions

ZekeAranyLucas:
Fixes #1420 and #1393 

1420 + 1393 - download projects as CSV and PDF
    * add ProjectTableRow component
    * dynamically load rules on init
    * enable download CSVs on menu with CSVLink
    * enable print PDF on menu with PdfPrint
    * switch cursor to hand/pointer on menu
    * update trash and copy icons to use SVG (pretty!)
  Tests 
    *  add ProjectRowTable tests for conditional state
    * updated svg icons to use title for testing and a11y
    * all tests need "await waitFor"
    * fixed tests for tdm-engine, NavBar, Level0Page
    * fixed tests for postPublicComment
    * skipped tests for PublicComentPage

TODO: decide about whether to use codes or names for column values in CSV file.
we should check it out in practice and then make a call.

check issues for details.

no screenshots, but the icons do look better.  probably need to add hover effects to the menu, too.


ZekeAranyLucas:
Hold off merging. Looks like I introduced a regression.  

when navigating from "Create Project" to "My Projects" this toast error:
```
ERROR
ResizeObserver loop completed with undelivered notifications.
    at handleError (http://localhost:3000/static/js/bundle.js:128757:58)
    at http://localhost:3000/static/js/bundle.js:128776:7
```

doesn't repro on refresh of "MyProjects" or navigation from "About".

entrotech (approved):
This is awesome! I'll create a follow-up issue about re-arranging some of the columns on the CSV format - we want to include some information from the product object, and omit columns that are intermediate calculations, and order the columns in roughly the order they appear in the calculation wizard.  We'll assign it to someone else.

Issue #1238, opened 2022-08-25
Implement Server-Side Request Validation and API endpoint tests


entrotech:
### Dependency: 
**This issue tracks all sub issues below. It can be closed upon completion of all sub issues**

### Overview
A best practice is for the Web API server to verify that the body of all requests meet appropriate validation criteria. We need to create a workflow for testing all endpoints. We will need to set up tests and web API request validation for the endpoints.   
Each route must complete the following:  
- Identify overexposed public surface and delete (i.e. remove any unnecessary/stale code as there are may be endpoints for features that will never be used)
- Write tests for all endpoints for a given route. Confirm there is documentation for testing
- Server side validation schema must be set up for the endpoints

### Action Items  
Routes:  
- [x] #1400
- [x] #1499
- [x] #1554
- [x] #1581
- [x] #1625
- [x] #1624
- [x] #1629

### Resources/Instructions
Server Side Validation
- https://www.npmjs.com/package/express-json-validator-middleware
- https://simonplend.com/how-to-handle-request-validation-in-your-express-api/  

Testing
- [TDM Testing Wiki](https://github.com/hackforla/tdm-calculator/wiki/Testing#testing-the-tdm-server)
- https://jestjs.io/
- https://node.testcontainers.org/

mojimoh:
@entrotech does not having this present a security risk for the site?

entrotech:
@agosmou @azajzon   This is partially implemented for the /accounts routes in the develop branch at this time. See the referenced article in this issue description for the explanation.  The middleware for validation is in the /server/middleware/validate.js file, the example schemas in the /server/app/schemas folder, and  these are wired to the /accounts/* endpoints at the end of the accounts controller /server/app/controllers/account-controller.js.

To address concerns raised by Issue #1360 , we might want to make validation for the registration endpoint more restrictive - e..g. limit to valid-looking emails, and first and last names that aren't obvious script injection attempts.

entrotech:
All component issues completed!

Issue #46, opened 2019-08-06
An Admin should be able to create and maintain calculations


entrotech:
### Dependencies
- [x] After all the issues in the prioritized backlog have milestones, release dependency and prioritize this. 

### Overview
LADOT would like a feature where they can perform CRUD operations on strategies so that they can keep a tool current with the regulation.

### Action Items
- [ ] Define requirements (e.g. what happens to saved projects when a strategy change is made)
   - [ ] Auditability 
   - [ ] User Notifications
   - [ ] Administrator Screen
   - [ ] Edge Case Scenarios

#### Tables involved
This is a CRUD feature against the Calculation table in the db. 

#### Who would use it and why
It would be used by an admin to create and maintain different types of calculations (maybe TDM for other cities with different rules, test calculations or calculations for completely different purposes, such as carbon tax credits, property tax rates, etc.).

agosmou:
This is stale and will not be implemented per [comment](https://github.com/hackforla/tdm-calculator/pull/1521#issuecomment-1820080783)

PR #1521, opened 2023-11-17
Calculation Endpoints: Adds Validation and Tests 

### LINKED ISSUE #1499: Engineer tests and validation for 'Calculation' Web API requests
### Overview

We need to create a workflow for testing all 'calculation' endpoints. We will need to set up tests and web API request validation for the endpoints.

### Details | Scope of Work
We need to test the API endpoints using 'jest'  

### Action Items

- [x] write all tests for endpoints and documentation
- [x] set up server side validation schema

### Resources/Instructions
Server Side Validation
- https://www.npmjs.com/package/express-json-validator-middleware
- https://simonplend.com/how-to-handle-request-validation-in-your-express-api/  

Testing
- https://jestjs.io/
- https://testcontainers.com/

agosmou:
Fixes #1499 

Sub-Issue from #1238 

- Added integration tests for the 'calculation' endpoints that test the functional requirements of the calculation workflow as well as edge cases
- No need for validation as overexposed endpoints were deleted

- Calculation Workflow Quality: Added tests for 'calculation' endpoints to ensure workflow meets functional requirements and handles edge cases
- Enhanced CI/CD: Integration tests in TDM CI pipeline prevent PRs that merge breaking changes
- Reliable Testing: Integrated 'Jest' and 'testcontainers' to add robust testing reflecting our production setup

### Notes
See [Test Wiki](https://github.com/hackforla/tdm-calculator/wiki/Testing#testing-the-tdm-server) for testing documentation on TDM Calculator


server/.env.example
agosmou, line 1: I changed the file name from 
.example.env -> .env.example  as it was before (better naming)

server/app/services/calculation.service.js
```diff
@@ -38,7 +38,7 @@ const post = async item => {
     request.output("id", mssql.Int, null);
     const response = await request.execute("Calculation_Insert");
 
-    return response.outputParameters;
+    return response.output;
```
agosmou, line 41: This was returning an undefined variable so I used the VSCode debugger to check what the property was supposed to be - it's been rectified now and actually returns the `id`

server/__tests__/calculation.test.js
```diff
@@ -0,0 +1,122 @@
...
+        name: "New Test Name",
+        description: "New Test Description",
+        deprecated: true,
+        id: newCalcId
+      });
+    expect(res.statusCode).toEqual(200);
+  });
+
+  // PUT "/:id" Update calculation using inexistent id (Admin only)
+  //TODO: this endpoint logic needs error handling for inexistent ids
```
agosmou, line 110: Consider adding issue - inexistent id's also returned 200 status codes

server/__tests__/calculation.test.js
```diff
@@ -0,0 +1,122 @@
...
+  // DELETE "/:id" Delete a calculation (Admin only)
+  it("should delete a calculation", async () => {
+    const res = await request(server)
+      .delete(`/api/calculations/${newCalcId}`)
+      .set("Authorization", `Bearer ${adminToken}`);
+    expect(res.statusCode).toEqual(200);
+  });
+
+  // DELETE "/:id" Delete a calculation using inexistent id (Admin only)
+  //TODO: this endpoint needs error handling for inexistent ids
```
agosmou, line 121: Consider adding issue - inexistent id's also returned 200 status codes

server/app/services/calculation.service.js
ZekeAranyLucas: Nice!

server/__tests__/calculation.test.js
ZekeAranyLucas: Agreed.

server/__tests__/calculation.test.js
ZekeAranyLucas: Agreed

server/.env.example
```diff
@@ -67,31 +67,33 @@ APPLICATIONINSIGHTS_CONNECTION_STRING=
...
-# CLIENT_URL=
-# SERVER_URL=
+# CLIENT_URL=http://localhost:3001
+# SERVER_URL=http://localhost:5002
 
-# SENDGRID_API_KEY=
-# EMAIL_SENDER=
-# EMAIL_PUBLIC_COMMENT_LA_CITY=
-# EMAIL_PUBLIC_COMMENT_WEB_TEAM=
+# SENDGRID_API_KEY=SG.testAPIkey
```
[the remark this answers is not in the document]
ZekeAranyLucas: Does this mean tdm-server-tests can run with .env.example?

server/app/schemas/calculation.post.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  type: "object",
+  required: ["name", "description", "deprecated"],
+  properties: {
+    name: {
+      type: "string",
+      minLength: 1,
+      pattern: "^[a-zA-Z '.-]+$"
```
ZekeAranyLucas, line 8: What will happen in the UX if the user breaks validation?

server/__tests__/calculation.test.js
```diff
@@ -0,0 +1,122 @@
...
+      .post("/api/calculations")
+      .set("Authorization", `Bearer ${adminToken}`)
+      .send({
+        name: "Test Name",
+        description: "Test Description",
+        deprecated: false
+      });
+    expect(res.statusCode).toEqual(201);
+    expect(res.body).toHaveProperty("id");
+    newCalcId = res.body.id;
```
ZekeAranyLucas, line 92: expect()?

agosmou: line 92 intentional - this sets the variable that is used in the following tests

server/app/schemas/calculation.post.js
agosmou: I think the create, update , delete, calculation endpoints are [not used](https://github.com/search?q=repo%3Ahackforla%2Ftdm-calculator+calculation+path%3A%2F%5Eclient%5C%2Fsrc%5C%2F%2F&type=code&p=1) by the client currently - future implementation. I think we mostly use the GET endpoints. Also, there is only '1' calculation in the db currently.

server/.env.example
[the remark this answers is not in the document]
agosmou: Nope! I just changed it because VSCode has better highlighting with the new name. Also included the variables so new members could run the tests easily without the formal variables. I tried to get this to work unsuccessfully with `./push.sh` 
An alternate strategy I thought of this morning is a function in the push script that can uncomment the test variables for the server tests and then re-comment upon completion. I'll have to check if this is viable, but editing the .env file with code feels unkosher. Thoughts?

server/app/schemas/calculation.post.js
ZekeAranyLucas: Then how do you know if this validation is right?

server/__tests__/calculation.test.js
ZekeAranyLucas: Other tests depend on this side effect? Tests need to runnable as a single unit.

server/app/schemas/calculation.post.js
ZekeAranyLucas: The answer could be you are making stubs, but then the code needs a comment that the regex are not spec'ed. 

I am looking for answers to why in the code.

ZekeAranyLucas: And unused endpoints are risky. I am inclined to delete dead code as a security risk. 

Is there a plan for when to add it?

server/.env.example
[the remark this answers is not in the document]
agosmou: I was looking into that but had some concerns.

Since the container has a single database that is [restored before every test suite](https://github.com/hackforla/tdm-calculator/blob/develop/server/_jest-setup_/local-setup/setup-after-env.js) (i.e. before every test file), the logic that handles this will call `tedious-pool.js` the [connection logic](https://github.com/hackforla/tdm-calculator/blob/develop/server/app/services/tedious-pool.js) from the server which will consume the original .env file. this means that this file would require a conditional to see which `.env` to consume right? Is this crufty?

server/__tests__/calculation.test.js
agosmou: It's intentional for the nature of these tests since it avoids mocking - everything tries to be as "real" as possible. We discussed adding functional tests that would be as close to the production environment as possible. So certain tests have protected endpoints which require JWT's

We had discussed trying to add unit tests but this way added more coverage and better ROI on time for where we are in the project.

Thoughts?

server/app/schemas/calculation.post.js
agosmou: excellent question. Ill have to refer to @entrotech

not to mention validation logic may even require numbers to be valid (not included currently)

agosmou:
> Looking good. What prompted adding the schema validation?

@ZekeAranyLucas 

Request from John to add validation for all web api endpoints - #1238

server/.env.example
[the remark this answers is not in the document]
ZekeAranyLucas: It means everywhere we use dotenv should probably go through a tdmenv wrapper.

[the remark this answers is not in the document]
agosmou: so new issue for refactoring codebase with this ^

server/__tests__/calculation.test.js
ZekeAranyLucas: For example, jest can rerun just failed tests or based on regex. As your tests scale, running subsets and reordering will become even more important. 

Don't let tests depend on other tests.

ZekeAranyLucas (changes requested):
In the name of security, I recommend deleting unused service  methods. This change becomes much smaller. 😀

server/__tests__/calculation.test.js
agosmou: > If the tests must run in a certain order they should just be a single test.
I put it in a single 'describe' block with this intention - is that pattern inappropriate? 

> Don't let tests depend on other tests.
Could this be done without mocking?

ZekeAranyLucas:
> In the name of security, I recommend deleting unused service  methods. This change becomes much smaller. 😀 

After looking at #1238, I have better context. Write tests for the faq crud APIs first. Those ones are accessible in the client, and need the attention.

ZekeAranyLucas:
Ok, editing calculations is covered by issuer #46. Given the nature of how rules work, I suggest only devs should ever change rules. They are too finicky for normal users to get right. Allowing an admin to delete a rule is a recipe for disaster

I recommend:

* resolve 46 as "won't fix"
* delete related service methods
* use #1238 issue to identify similar overexposed public surface and delete.

What do you think @entrotech ?

.github/workflows/tdm-server-tests.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - develop
 jobs:
-  build:
+  test:
```
agosmou, line 8: better naming convention for the workflow

server/__tests__/calculation.test.js
ZekeAranyLucas: Describe is any arbitrary semantic group. "It" is the test. 

So you could group tests by all needing the same before/after assumptions. 

```
describe ('tests that need an calculation')
  let calcId = null
  beforeEach() { calcId = ... }
  afterEach() { cleanup(calcId)}
  it('verifies delete') { 
    delete(calcId);
    expect(get(calcId)).toBeFalsy()
  }
  it('verifiers update') ...
```

So each test starts with clean assumptions.

agosmou: Just refactored based on this suggestion

commit 7f1deff8af2698b17763ceb17ab8f937677fc188

ZekeAranyLucas:
@agosmou is this ready to roll? are you going to create issues after merging?

entrotech:
@ZekeAranyLucas  comments above are spot-on. Early on, we had thoughts of implemented admin screens for versioning the calculations, allowing a super-admin to edit rules on an existing calculation to create a new version, and apply that version to future projects, but the implications are kinda scary, and I like having enough friction in changing rules that they are made mostly by DB Migrations and an occasional code change, so they need to go through a QA process by people that know what they are doing.

ZekeAranyLucas (approved):

agosmou:
The ['rule service'](https://github.com/hackforla/tdm-calculator/blob/develop/client/src/services/rule.service.js#L16) uses the getById service from the 'calculations' server service so this will remain for now.

agosmou:
The above requested changes were addressed in the latest commit [f1dcd6d](https://github.com/hackforla/tdm-calculator/pull/1521/commits/f1dcd6d232b5387b00f7983914222fa2e4c0e9b9)

entrotech (approved):
I trust you two to have worked out the details on this PR. I am agreeing on deleting the endpoints you mentioned.

BTW, there is a remnant of Issue #46 that I can't bring myself to remove, just cause it's kinda cool.  Log in as an admin and change the client path to /admin and it give you an interesting view of the rules that make up the calculation.

@agosmou you can merge this yourself, if you want.

ZekeAranyLucas:
/admin is cool!  i wouldn't remove that.

agosmou:
Yes! Very cool (see below for record). This is unaffected by this PR.  

---
image

Issue #1157, opened 2022-05-17
Users could not find Checklist once closed


SMR9220:
### Overview
Users need to be able to find the checklist once it is closed.

### Action Items
-  [ ] Provide a clickthrough orientation of where to find things upon landing, similar to the popovers used for new product releases.

OR

- [x] Put checklist in a top level navigation help drop down.   add animation showing the checklist shrinking to the corner

### Resources/Instructions
Some users were worried that the checklist may not have what they needed to use the calculator, but they proceeded anyway. They could not find the checklist after it was closed.

jahpe777:
Fixed issue of Checklist Modal not showing up initially on every page. All Checklist links within Footer should work on every page now.

jahpe777:
Added animation for closing the Checklist modal so that the user will have a visual to where the Checklist button is if they want to see it again.

Biuwa:
https://user-images.githubusercontent.com/98936028/183279098-741379a3-ab74-4746-82da-e04039036cb2.mov

ExperimentsInHonesty:
showed it to stakeholders, we agree it should be a little slower.

Biuwa:
@jahpe777 During today's meeting, stakeholders sought to slow down the checklist closing animation as the speed was too fast.

jahpe777:
slowed down speed of animation from 1000ms to 1500ms

Biuwa:
Slowed down checklist closing Animation 

https://user-images.githubusercontent.com/98936028/185002128-3917adba-aec4-401d-a5b2-47c3d6c81ce1.mov

Issue #1086, opened 2022-03-10
Checklist popup for the create project page


seenaiype:
### Overview
To provide some guidance on the information users that would be helpful for the calculator, we are going have a pop-up show with the list of required information and information around exemption 

### Action Items
- [x] Popup to show to all users doing a new project
- [x] popup shall showup only on the first page
- [x] add a link to the footer of the page, for easy access to the popup 
- [x] Popup for checklist shall be visible after the terms and condition
- [x] popup shall show for both logged in and not logged users
- [ ] To be reviewed with the product team
- [ ] To be reviewed with the stakeholders
- [x] Add wording  "Check LAMC Section 12.26 J.3(b)(3) of the **Draft Revised TDM Ordinance** for exemption details."
       - 

### Resources/Instructions
[Figma](https://www.figma.com/file/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?node-id=5683%3A20479)
#693 - terms and condition story, to make sure we do this in co-ordination

jahpe777:
@seenaiype @KPHowley - Should the modal close simply by clicking the "X" **only** or should it **also close** if you click outside of the modal?

1. Progress: I have the basic structure in place. Finalizing styling. Click events need to be completed. Local storage needs to be completed
2. Blockers: May want help with events
3. Availability: Somewhat available
4. ETA: March 30th

jahpe777:
Finalized checklist modal 
![Screen Shot 2022-03-20 at 2 22 37 PM](https://user-images.githubusercontent.com/40730303/159186525-0e3c059c-cc53-4c85-a3bc-5cda17a3c68b.png)

jahpe777:
Updated styling and fixed relationship between Terms and Conditions modal and Checklist modal

PR #1526, opened 2023-11-20
1478 - fixes checklist link from faq page

### LINKED ISSUE #1478: Fixing the Checklist Hyperlink on the FAQ page
### Overview
We need to fix the checklist hyperlink on the FAQ page, so users can see the checklist pop up when they click on it

### Action Items
- [x] Create a new page on the site called "checklist".  Do not add it to any menus.  
- [x] Once created link the page to the word checklist in the FAQ question "What do I need to prepare to get started?"

### Resources/Instructions
https://tdm-dev.azurewebsites.net/faqs

ZekeAranyLucas:
Fixes #1478 by switching the checklist content from a modal prompt to a navigable page.
fix 1478 by making ChecklistPage
    * added ChecklistPage including routes in App
    * added redirect for "faqs/true" which is in DB
    * updated ModalDialog to have title for a11y and testing
    * simplified the terms and conditions state management
    * clicking question on FAQ now intutively expands answer
    * fixed/skipped broken client tests
    * fixed hide/unhide in ProjectContextMenu

checklist modal was not navigable,

see also slack: https://hackforla.slack.com/archives/CKY65G266/p1700149562824069

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -260,3 +260,5 @@ dotenv
 yarn.lock
 
 .mssql.json
+server/.env-ale
+server/.env-alpha
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -12,6 +12,8 @@ import Footer from "./components/Footer";
 import About from "./components/About";
 import TermsAndConditionsPage from "./components/TermsAndConditions/TermsAndConditionsPage";
 import TermsAndConditionsModal from "./components/TermsAndConditions/TermsAndConditionsModal";
+import ChecklistModal from "./components/Checklist/ChecklistModal";
+import ChecklistPage from "./components/Checklist/ChecklistPage";
 import PrivacyPolicy from "./components/PrivacyPolicy";
 import Register from "./components/Authorization/Register";
 import UpdateAccount from "./components/Authorization/UpdateAccount";
@@ -28,7 +30,6 @@ import ForgotPassword from "./components/Authorization/ForgotPassword";
 import "./styles/App.scss";
 import Feedback from "./components/Feedback/FeedbackPage";
 import Sidebar from "./components/Sidebar";
-import ChecklistPage from "./components/Checklist/ChecklistPage";
 
 const useStyles = createUseStyles({
   app: {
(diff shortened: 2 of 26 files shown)


entrotech (changes requested):
In an effort to be able to access the checklist from an answer in the FAQs, we've completely broken the primary use case for the checklist.  

The primary purpose of the checklist is for users in the Wizard to have a way to get a list of the materials and information they need to successfully complete the wizard. At first, we put the checklist directly on the first page of the Wizard, but then decided that it would be better as a modal that is automatically shown when the user starts the Wizard, so they won't miss it (Issue #1086). This was pretty good, but we wanted them to be able to re-open the wizard from any page, so added the link in the toolbar (Issue #1157) that would allow them to re-open the dialog from any page in the wizard, and (later) anywhere in the site for reference purposes.  When using the wizard it is important that we do **not** navigate away from the wizard, because they will lose any work they have done on their calculation. We also added an animation that took 1.5 seconds to close the modal and showed it shrinking and disappearing into the footer link, so users would notice the link, and deduce that the link could be used to re-open the checklist if they needed it again (See comments on Issue #1157). 

Normally, if you try to leave the wizard without saving your changes, you should see the NavConfirmModal, asking you if you really want to leave the wizard and lose your changes - but i think this got disconnected about two weeks ago when I upgraded the react router and didn't get this part working again. We need to get this working again, and I'll create an issue for it.

However, if we fix this and leave the checklist implementation as it is in this PR, a user on the wizard with a partially-completed calculation will not be able to access the checklist at  all without losing their partially completed calculation or saving it.

Since there a number of other changes included in this PR that are good steps forward, we should probably just try to start restoring the checklist functionality by having the footer Checklist link open the checklist as a dialog without navigating the router. I'll create separate issues to hookup the NavConfirmModal again and re-implement the animation when the confirm dialog is closed.

ZekeAranyLucas:
> In an effort to be able to access the checklist from an answer in the FAQs, we've completely broken the primary use case for the checklist.
> ...
> Since there a number of other changes included in this PR that are good steps forward, we should probably just try to start restoring the checklist functionality by having the footer Checklist link open the checklist as a dialog without navigating the router. I'll create separate issues to hookup the NavConfirmModal again and re-implement the animation when the confirm dialog is closed.

Thank you, John, that's exactly what I was hoping to understand. I think that a collapsible section would be more usable, but for now I will replace the modal behavior.

ZekeAranyLucas:
latest commit reverts to old user behavior.  first run gets a double modal prompt with terms and checklist.  footer link shows the modal. faq still shows a separate page.

Issue #1516, opened 2023-11-16
Set Project Page to auto-refresh after Project actions


Jonathanko52:
### Overview
On the Project page, after deleting, duplicating, or converting a project to a snapshot, the page doesn't refresh to reflect these changes. It can appear as if the task failed, and might lead to a user attempting to create addition duplicates, delete non-existent projects, and so forth.

### Action Items
1.Set Project Page to refresh after there actions.

PR #1528, opened 2023-11-21
1518 + 1516 after context menu action, close menu and update project row

### LINKED ISSUE #1518: My Project Page Context Menu Auto-Close and refresh grid
### Overview
The Context menu on each row of the My Projects Page is working ok, except that when you select the various actions, completion or cancellation of the action should close the context menu (which currently happens only for action that have a confirmation dialog), and refresh the grid contents by re-querying the database.

### Action Items
- [ ] Fix

### Resources/Instructions
image

ZekeAranyLucas:
Fixes #1518 + #1516 together

* close menu explicitly by wrapping the action handlers
* refresh projects collection after making service calls

### DIFF
--- a/client/src/components/Projects/ProjectsPage.js
+++ b/client/src/components/Projects/ProjectsPage.js
@@ -101,7 +101,7 @@ const ProjectsPage = ({ account, contentContainerRef }) => {
   const email = account.email;
   const navigate = useNavigate();
   const handleError = useErrorHandler(email, navigate);
-  const projects = useProjects(handleError);
+  const [projects, setProjects] = useProjects(handleError);
   const [orderBy, setOrderBy] = useState("dateCreated");
   const [copyModalOpen, setCopyModalOpen] = useState(false);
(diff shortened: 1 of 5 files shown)


client/src/components/Projects/ProjectsPage.js
```diff
@@ -138,6 +138,10 @@ const ProjectsPage = ({ account, contentContainerRef }) => {
     setCopyModalOpen(true);
   };
 
+  const updateProjects = async () => {
```
entrotech, line 141: This seems sort of redundant with the useProjects hook, since both are calling the same projectService.get method. Maybe we should consider modifying the useProjects hook to support additional features like re-executing the query.  We have upcoming issues to filter projects by various criteria (hidden/visible, draft, snapshot, active/deleted, possibly by owner, etc.) which can probably be done front-end. Maybe the filtering is included in the hook?

entrotech (approved):

PR #1555, opened 2023-12-06
Added inline if-else to remove click function and hover pointer class

irisslee:
Fixes ##1552

- Added inline if else to remove click function (one that sorts) and hover pointer class 
-
-

- Because the last column shouldn't sort or change pointer on hover
-
-

<details>

image

</details>

<details>
  
image

</details>


client/src/components/Projects/ProjectsPage.js
```diff
@@ -471,8 +471,14 @@ const ProjectsPage = ({ account, contentContainerRef }) => {
                         return (
                           <td
                             key={i}
-                            className={`${classes.td} ${classes.theadLabel}`}
-                            onClick={() => handleSort(header.id)}
+                            className={
+                              i == 8
```
ZekeAranyLucas, line 475: what does 8 mean?

this feels fragile, the design folks keep changing the required columns. consider structuring it to make it clearer why this code is needed.

ZekeAranyLucas (changes requested):
nice getting this in a PR so fast! :)

client/src/components/Projects/ProjectsPage.js
irisslee: [https://github.com/hackforla/tdm-calculator/blob/8c0d6bc7fa3b80aa4a7f8b0fcef8a7e93bafd006/client/src/components/Projects/ProjectsPage.js#L366-L385](url)

The function iterates through this list of headers and adds classes and click functions. I'm just using the index on this list of headers. I agree that it's fragile! Another option would be to add conditional based on `label` being blank string, or maybe we can add a boolean property to this header list.... something like - `"sortfunction": true/false` and base the conditionals off of that. I prefer the latter - I think it's more explicit. What do you think?

irisslee:
@entrotech @ZekeAranyLucas  Here are my changes

ZekeAranyLucas (approved):
much clearer! ship it

PR #1563, opened 2024-01-04
1556 Rename Snapshot

### LINKED ISSUE #1556: Implement the "Rename Snapshot" feature
### Overview
User should be able to rename an existing snapshot from the context menu on the My Projects Page.

### Action Items
- [x] Create a Project_Rename stored procedure that is similar to the Project_Snapshot stored procedure, except that it does not update the dateSnapshotted column.
- [x] Create a web api endpoint for PUT to /api/projects/rename that will rename a project. The web api router should use the jwtSession.valdiateUser middleware to verify that the request is from a logged-in user, The request should have the same body as the /api/projects/snapshot endpoint.
- [x] Create a client-side service that makes the PUT request described above.
- [x] Modify the context menu to display the "Rename Snapshot menu item in the context menu as shown in the below screenshot when
  - [x] The project is a snapshot (dateSnapshotted is not null)
  - [x] The project is owned by the logged-in user
  - [x] If the project is not a snapshot, the existing "Convert To Snapshot menu Item should be shown instead.
  - [x] If the project is a snapshot, but not owned by the logged-in user, the menu item should be disabled (greyed-out and not responsive to click events.
  - [ ] Create a dialog as shown below that prompts the user to supply the new project name. When the user clicks "Rename", the dialog will close and the client service called to effect the snapshot re-naming. You should consult with the UI/UX design team for the exact wording on the dialog, but use the existing SnapshotModalDialog.js in the code for the implementation code that should be used, so we use our "standard" modal dialog design.

### Resources/Instructions
image
image

Jonathanko52:
Fixes #1556

1.Set up client side put service
2. api endpoint.
3.Project_Rename stored procedure
4. Snapshot rename modal
5. Edited context modal to conditionally change between "convert to snapshot, rename snapshot, rename snapshot(disabled).

1. Allow users to rename snapshots they've created.
2. NOT-allow users to rename other peopler's snapshots.

### DIFF
--- a/client/src/components/Projects/ProjectContextMenu.js
+++ b/client/src/components/Projects/ProjectContextMenu.js
@@ -10,7 +10,8 @@ import {
   faCamera,
   faTrash,
   faClone,
-  faFileCsv
+  faFileCsv,
+  faPencil
 } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 
@@ -31,6 +32,12 @@ const useStyles = createUseStyles({
       cursor: "pointer"
     }
   },
+  listItemDisabled: {
+    display: "flex",
+    flexDirection: "row",
+    padding: "0.5rem",
+    color: "gray"
+  },
   listItemIcon: { marginRight: "0.3rem" }
 });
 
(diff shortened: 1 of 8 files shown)


entrotech (approved):
I will create a PR with the script for the Project_Rename stored procedure

PR #1583, opened 2024-01-20
Logout Functionality

### LINKED ISSUE #412: Delete JWT on Logout
### Overview
Investigate best practice for removing JWT cookie on Logout. It currently remains after logging out, which could create security vulnerability.

### Action Items
- [x] RESEARCH: Read on best practice for storing jwt info. We currently use httpOnly Cookie. Is localStorage better? Is httpOnly necessary?
#### JWT Storage Best Practices
[Use cookies](https://www.loginradius.com/blog/engineering/guest-post/jwt-authentication-best-practices-and-when-to-use/#:~:text=The%20easiest%20ways%20to%20store,to%20store%20JWT%20in%20cookies) 
#### What is `HttpOnly` prop of cookie
  - In short, the HttpOnly flag makes cookies inaccessible to client-side scripts, like JavaScript. Those cookies can only be edited by a server that processes the request.
  - So long as you are not running javascript to collect site traffic data or analytics, then you can use HttpOnly to provide an extra layer of protection for your website. 
  - HttpOnly tags are especially useful on websites that — for one reason or another — have not yet adopted the HTTPS protocol.
  - When not to use HttpOnly
    - Using javascript for analytics
    - Using javascript for page state. If your website's page state depends on javascript to interpret and modify your cookie values, this site type of site architecture may not work with HttpOnly and could leave the site open to cross-site request forgeries (CSRFs). In cases like this, you should leave out the HttpOnly flag and simply use the secure flag with the HTTPS protocol.
  - A cookie with the HttpOnly attribute is inaccessible to the JavaScript `Document.cookie` API; it's only sent to the server.

##### Cookies
- Cookies in JavaScript are used to read from and store small data related to the user on the browser.
- Local Storage allows you to also read from and store data related to the user on the browser using JavaScript's [Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage).
- An example of using cookies is storing the user's authentication token when they are logged in. The server needs to have access to that cookie to authenticate the user in later requests and retrieve data or process actions based on who the user is.
- Cookies on the client-side (the browser) are natively accessed with document.cookies. When you store cookies on the browser, you can make them specific to a path in a domain name and you can add an expiry date for them.
    - Except HttpOnly flag

##### Local Storage
- Local Storage is used to read and write data in the browser's storage. The data is persistent and has no expiry date.
- An example of using local storage is storing the user's color mode preference (light or dark mode). This piece of information is generally not necessary for the server to know, and is used to ensure that the user's preference is reflected on the website.

##### The Difference
- The server can access the client's cookies but not the data stored in the client's storage.
- So, in use cases where it's important for the server to have access to a set of data, you should use cookies.
- Cookies should not be used to store large pieces of data. For example, if you want to store the user's details in the browser then it's best to store them in the local storage.

##### Which to use
- If you need to store data that is accessible for both the server and the client, use cookies. Otherwise, use local storage.
- If you need to store larger data, use local storage.
- If you need to store data that does not expire, use local storage.
- If you need easy-to-use methods to access and modify the data stored on the client, use local storage.

~~- [ ] APPLY: I anticipate the solution will either be:~~
  ~~- converting the cookie to not httpOnly and deleting it from client on logout.~~
  ~~- or creating a server endpoint to remove the cookie~~
#### Technical Solution
- [ ] Understand Login and Logout flow
   - ..routes/account.routes
   - Here we have login and logout routes.
   - Note that I am able to log to console during login, but logout route is not hit on logout.
     - Is logout handled another way? Possibly by the account.controller's login method?
- [ ] As there is no way to delete a jwt, we need to set the expiration to a date shortly after logout. Currently, we set this expiration to 12 hours at login. At logout, we can get the token from the `request`, and then `set` the token with a new expiration dateTime.
- [ ] UPDATE DOCUMENTATION: update register.md which currently indicates jwt should be deleted from client.

### Previous related issues:

- Issue #400 
- Issue #71 
- Issue #48 
-

neyaadeez:
Created Logout File in client/src/components/Authorization/Logout.js, JWT Token - Clear cookie

Fixes #412 

- Created a logout route client/src/App.js
- Added a new file for logout functionality: client/src/components/Authorization/Logout.js
- Edited client/src/components/Layout/NavBarLogin.js - Changed pathname: /login/${(account && account.email) || ""} to pathname: /logout`,
- Added logout function in service file: client/src/services/account.service.js
- Modified logout function in server route: server/app/routes/account.routes.js to clear the JWT cookie.

- To resolve jwt token issue

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -37,6 +37,7 @@ import ResetPassword from "./components/Authorization/ResetPassword";
 import ForgotPassword from "./components/Authorization/ForgotPassword";
 import Feedback from "./components/Feedback/FeedbackPage";
 import ErrorPage from "./components/ErrorPage";
+import Logout from "./components/Authorization/Logout";
 
 const calculationPath = "/calculation/:page/:projectId?/*";
 
@@ -138,6 +139,7 @@ const App = ({
           <Route path="/updateaccount/:email?" element={<UpdateAccount />} />
           <Route path="/confirm/:token?" element={<ConfirmEmail />} />
           <Route path="/login/:email?" element={<Login />} />
+          <Route path="/logout" element={<Logout />} />
           <Route path="/forgotpassword" element={<ForgotPassword />} />
           <Route path="/resetPassword/:token" element={<ResetPassword />} />
           <Route
--- a/client/src/components/Authorization/Logout.js
+++ b/client/src/components/Authorization/Logout.js
@@ -1,0 +1,18 @@
+import React from "react";
+import { useNavigate } from "react-router-dom";
+import * as accountService from "../../services/account.service";
+
+const Logout = () => {
+  const navigate = useNavigate();
+  const handleLogout = async () => {
+    await accountService.logout();
+    // Redirect to the "/login" page after the logout
+    navigate("/login");
+  };
+  // Call handleLogout when the component mounts
+  handleLogout();
+  // You can return some UI if needed (e.g., a loading spinner)
+  return <div>Logging out...</div>;
+};
+
+export default Logout;
--- a/client/src/components/Layout/NavBarLogin.js
+++ b/client/src/components/Layout/NavBarLogin.js
@@ -54,7 +54,7 @@ const NavBarLogin = ({ classes, handleHamburgerMenuClick }) => {
       <Link
         className={`${classes.link} ${classes.lastItem}`}
         to={{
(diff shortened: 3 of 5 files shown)


entrotech (approved):
Just based on a look at the code, it seems reasonable. Good idea to create a <Logout> component.  I'm going to just go ahead and merge it.

PR #1584, opened 2024-01-22
Fix date format/filter on filter drawer

### LINKED ISSUE #1575: Fix Date Filters on My Projects Page
### Overview
On the My Projects page, dates are formatted as YYYY-MM-DD on the grid, but as MM/DD/YYYY on the Filter drawer. Bonnie wants the filter drawer to be changed to use the date format YYYY-MM-DD.
Also, the date filters do not seem to be applied at all.

### Action Items
- [ ] fix date format in date pickers on filter drawer.
- [ ] fix filtering to filter by create or modified date ranges

mark-mxwl:
Fixes #1575

- Added dateFormat property to DateRangePicker component with YYYY-MM-DD format.
- Added conditional statements to filterProjects function to allow for date filtering.
- Created getTimeInMs helper function to convert dates to numbers for easier comparison.

- To resolve the noted filter drawer issues on the My Projects page.


client/src/components/UI/DateRangePicker.jsx
```diff
@@ -18,6 +18,7 @@ const DateRangePicker = ({ startDate, endDate, setStartDate, setEndDate }) => {
         startDate={startDate}
         endDate={endDate}
         customInput={<DatePickerCustomInput />}
+        dateFormat="yyyy/MM/dd"
```
entrotech, line 21: Can the format use dashes like yyyy-MM-dd so the punctuation matches the date columns in the grid?

client/src/components/Projects/ProjectsPage.js
```diff
@@ -355,6 +359,27 @@ const ProjectsPage = ({ account, contentContainerRef }) => {
     )
       return false;
 
+    if (
```
entrotech, line 362: The way this function is written, each conditional should test whether the project should be excluded, and return false if it should be excluded.  So you can't return true until all of the filter conditions are met, or you would be ignoring all the subsequent conditional tests.  So, for example, lines 367-371 should be  of "inverted" to return false if there is an endDateCreated and the project's dateCreated is after the endDateCreated.  Likewise for the lines 377-381

entrotech (changes requested):
Looks pretty close - See my comments on specific lines.

client/src/components/Projects/ProjectsPage.js
mark-mxwl: Thanks for the notes, @entrotech! I'll make the suggested changes and get this back up shortly.

client/src/components/Projects/ProjectsPage.js
```diff
@@ -337,8 +337,9 @@ const ProjectsPage = ({ account, contentContainerRef }) => {
     setFilterText(text);
   };
 
-  const getTimeInMs = date => {
-    return new Date(date).getTime();
+  const getDateOnly = date => {
```
mark-mxwl, line 340: In order to keep with filterProject’s convention of checking for false conditions, I had to adjust the getTimeInMs function to remove timestamps. While the project date entries all have timestamp values, the DateRangePicker component defaults all timestamps to 00:00:00. This creates issues, particularly when comparing “equal” dates: For example, Oct. 3rd at 5pm is “greater than” Oct. 3rd at 12am, thus returning false and omitting all end-date entries greater than 00:00:00. (This isn’t an issue with start-date entries because all selected timestamps are “less than” the project timestamps). For this scenario, the most reliable option I’ve found is to convert the date to a string, which omits the timestamp, then convert that string into a Date object for comparison.

client/src/components/UI/DateRangePicker.jsx
mark-mxwl: They can! Easy fix.

entrotech (approved):

PR #1654, opened 2024-04-04
1635 Clear trash projects quick fix

### LINKED ISSUE #1635: Implement the permanent delete after 90 days for projects.
### Overview
When deleting a project, the project should remain in the database(recycling bin or something) before it's permanently deleted in 90 days.

### Action Items
- [ ] Implement the permanent delete function (when someone checks table, send a call to check for a 'to be deleted flag', check the date, and delete it if it's been 90 days ).
- [ ] Add the 90 days message to the delete dialog.

### Resources/Instructions
- https://github.com/hackforla/tdm-calculator/issues/1383

**Additional context**
- Check to see if we already have a "recycling bin" in place.

Jonathanko52:
Fixes #1635 

When users view their saved projects, the fetch function now includes an imbedded delete function that scans through the fetched projects and deletes any that are in the trash and 30 days old.

The idea is for TDM to automatically delete all projects in the database that have been in the trash for 30 days. For now, we don't have that built into the back end, so we have the users automatically check and clean their own trash bin whenever they look at their projects. This does present a problem is the user trashes a bunch of projects and never logs in again, leaving their trash uncleaned, but this should work for now.

### DIFF
--- a/client/src/hooks/useGetProjects.js
+++ b/client/src/hooks/useGetProjects.js
@@ -10,6 +10,16 @@ const useProjects = handleError => {
       if (result.data === "" || result.data === false) {
         setProjects([]);
       } else {
+        console.log(result.data);
+        console.log(
+          "CURRENT",
+          (new Date() - new Date(result.data[0].dateCreated)) /
+            1000 /
+            60 /
+            60 /
+            24
+        );
+
         setProjects(result.data);
       }
     } catch (err) {


client/src/hooks/useGetProjects.js
```diff
@@ -5,12 +5,23 @@ const useProjects = handleError => {
...
       if (result.data === "" || result.data === false) {
-        setProjects([]);
+        setProjects(fetchedProjects);
       } else {
-        setProjects(result.data);
+        result.data.forEach(cur => {
+          let numberOfDaysSinceCreation =
+            Math.abs(new Date() - new Date(cur.dateCreated)) /
+            (1000 * 60 * 60 * 24);
+          if (numberOfDaysSinceCreation >= 30) {
```
entrotech, line 18: Issue says deletion should happen after 90 days, not 30.

client/src/hooks/useGetProjects.js
```diff
@@ -5,12 +5,23 @@ const useProjects = handleError => {
...
+            projectService.del(cur.id);
```
entrotech, line 19: This should be called async, i.e. `async projectService.del(cur.id);`

entrotech (changes requested):
The issue also calls for changing the message in the delete dialog to tell the user that the project will be permanently deleted after 90 days.

Jonathanko52:
Added text to delete modal, changed the date check to 90 days (also changed it to track "days since trashed" instead of "days since creation". Oops.) And extended the delete into its own separate, asynchornous function.

Screenshot 2024-05-01 at 7 25 07 PM

entrotech (approved):

PR #1692, opened 2024-05-07
Fix vulnerability in accountUpdate

### LINKED ISSUE #1691: Critical Security Alert #2389194
#### Describe the bug 

ITA reported a Critical Security Bug, which they all Alert # 2389194. This is a vulnerability in the Update Account feature, where a malicious user can create their own account, and modify a PUT request to the endpoint for updating an account to update a different users account.  This feature should be disabled until it can be better secured.

### Steps
- [ ] It looks like the endpoint is not secured properly, allowing a malicious user to impersonate a different user.
- [ ] Either fix the vulnerabilty or disable the feature until it can be secured properly.

### References

[Alert2486073.pdf](https://github.com/hackforla/tdm-calculator/files/15227579/Alert2486073.pdf)

entrotech:
Fixes #1691

- Modified url for updateAccount to PUT /api/accounts/updateAccount, instead of /api/accounts/:id/updateAccount, so
- service gets user's id from the secure token instead of the web api request url or body.

- See Issue #1691

### DIFF
--- a/client/src/components/Authorization/UpdateAccount.js
+++ b/client/src/components/Authorization/UpdateAccount.js
@@ -24,7 +24,6 @@ const UpdateAccount = props => {
   const classes = useStyles();
   const params = useParams();
   const initialValues = {
-    id: account.id || "",
     firstName: account.firstName || "",
     lastName: account.lastName || "",
     email: params.email || ""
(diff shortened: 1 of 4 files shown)


entrotech:
Replace with PR #1693

PR #1815, opened 2024-09-04
1646 visibility and status

### LINKED ISSUE #1646: Implement the Visibility and Status Column Filters on the My Project Page
### Overview
- Implement the Visibility and Status Column Filters on the My Projects Page per the Figma Prototype references in the Resources paragraph below to comply with the filter re-design of Issue #1578 
- 
### Action Items
- [x] The prototype uses toggle buttons for specifying the sort order for both of these filters, but this makes no sense, since the choices should be mutually exclusive. Check with designers @yiranshan to see what they want to do.
- [x] Implement the filtering for the Visibility column
- [x] Implement the filtering for the Status column
- [x] Also add the word "(deleted)" to the text in the Status column cells if the project is deleted, and make the row background the same color as shown in the prototype.  
- [x] The word "(deleted)" should no longer be appended to the Modified date.

### Resources/Instructions

Jonathanko52:
Fixes #1646

-Edited CSS for table header.
-Edited color change for header when a filter dropdown is selected
-Implemented filter functionality for Visibility Column
-Implemented filter functionality for Status Column
-Removed "deleted" from modified date
-Moved "delete" test to status column and changed the background for deleted rows.

Part of the revamp for the filtering.

Screenshot 2024-09-03 at 8 43 06 PM

Screenshot 2024-09-03 at 8 43 14 PM

Screenshot 2024-09-03 at 9 09 46 PM


client/src/components/Projects/ColumnHeaderPopups/ProjectTableColumnHeader.js
```diff
@@ -1,27 +1,54 @@
 import React from "react";
 import PropTypes from "prop-types";
 import "react-datepicker/dist/react-datepicker.css";
-import { MdFilterAlt } from "react-icons/md";
+import { MdFilterList } from "react-icons/md";
```
entrotech, line 4: You probably got the idea for this icon from the prototype, but, the prototype is several months old, and the design team has changed the icons they want to use in the design guide See https://docs.google.com/presentation/d/1I4q35NL2WW2RpksIyawhHCd8qJnrjme1ZDNIBxu24hQ/edit#slide=id.g2e77662f2ce_0_10  page 17.  The upshot is that you need to change the icon back to MdFilterAlt.

client/src/components/Projects/ColumnHeaderPopups/StatusPopup.js
```diff
@@ -4,6 +4,7 @@ import Button from "../../Button/Button";
 import RadioButton from "../../UI/RadioButton";
 import "react-datepicker/dist/react-datepicker.css";
 import { MdClose } from "react-icons/md";
+import UniversalSelect from "../../UI/UniversalSelect";
```
entrotech, line 7: I just figured out a day or two ago that the UniversalSelect control is not working.  If it is used as a controlled input, it does not not track the existing controlled value when it is first rendered.  We need to either fix it, or go back to using a standard react <select> control.

client/src/components/Projects/ColumnHeaderPopups/VisibilityPopup.js
```diff
@@ -4,6 +4,7 @@ import Button from "../../Button/Button";
 import RadioButton from "../../UI/RadioButton";
 import "react-datepicker/dist/react-datepicker.css";
 import { MdClose } from "react-icons/md";
+import UniversalSelect from "../../UI/UniversalSelect";
```
entrotech, line 7: See comment above about UniversalSelect not working.

client/src/components/Projects/ColumnHeaderPopups/VisibilityPopup.js
```diff
@@ -41,6 +49,9 @@ const VisibilityPopup = ({
     close();
   };
 
+  const handleChangeVisibility = visibilityValue => {
```
entrotech, line 52: Don't want to apply the change to the filter when the user picks a choice from the list - need to wait until they click on the Apply button.

client/src/components/Projects/ProjectTableRow.js
```diff
@@ -82,12 +82,7 @@ const ProjectTableRow = ({
   // deleted, in which case it will show the deleted date followed by "-Deleted" in red.
   const dateModifiedDisplay = () => {
     if (project.dateTrashed) {
-      return (
-        <span>
-          {formatDate(project.dateTrashed)}
-          <span style={{ color: "red" }}>-Deleted</span>
-        </span>
-      );
+      return <span>{formatDate(project.dateTrashed)}</span>;
```
entrotech, line 85: I don't know why you moved the deleted indicator to the visibility column. Maybe you misunderstood our conversation last week where I pointed out that the Date Modified column is where you would see that the deleted date is displayed if the project is deleted, per Emily's request, but we don't want to move it to the Status column.

client/src/components/Projects/ProjectsPage.js
```diff
@@ -733,16 +746,29 @@ const ProjectsPage = ({ contentContainerRef }) => {
                     <tr className={classes.tr}>
                       {headerData.map(header => {
                         return (
-                          <td key={header.id}>
+                          <td
```
entrotech, line 749: I guess what you are trying to do here is change the header appearance while it's corresponding popup is open.  This is outside the scope of this issue, and will be addressed by a separate issue, but if we wanted to do this, you don't need any new props at all. The reactjs-popup has a feature that allows you to change the trigger (i.e. Column Header appearance, depending on whether the popup is open or not - See the Trigger section on the page  https://react-popup.elazizi.com/component-api

client/src/components/Projects/ColumnHeaderPopups/StatusPopup.js
```diff
@@ -72,7 +118,10 @@ const StatusPopup = ({
         />
         <hr style={{ width: "100%" }} />
       </div>
-      <div>(Under Construction)</div>
+      <UniversalSelect
```
entrotech, line 121: You need two separate drop-downs on this popup. One controls criteria.status, which should have possible values ["active", "deleted", or "all"] the second controls criteria.type, which should have possible values of ["draft", "snapshot" or "all"]

entrotech (changes requested):
This PR is a lot more complicated than the issue calls for, and tries to implement changes outside the scope of the issue.  I'd like you to try again. This time: 

- Do not make changes to any components except VisibilityPopup and StatusPopup. For this issue, you should not need to add any props to either component.
- Make sure that filter and sorting criteria are not applied until the user clicks the Apply button.
- The Visibility Popup should use a set of three radio buttons to do specify the filtering as in the prototype, but you should omit the superfluous textbox found in the prototype.
- The Status Popup should be different from the prototype and use a set of three radio buttons to choose "Draft", "Snapshot or "Draft and Snapshot". There should be a checkbox like the prototype that is labelled "Deleted Projects". If it is checked, then show both active and deleted projects, if it unchecked, only show active projects.

client/src/components/Projects/ColumnHeaderPopups/StatusPopup.js
Jonathanko52: I think this was what resulted in me going out of scope for this task. I discovered that when I attempted to change the sorting option, it would reset the filter option. I believe that the issue was that the StatusPopup component was being rendered each time, and so the existing controlled value was lost. I decided the best approach to fixing this was to save that value in state of the Project page.

client/src/components/Projects/ProjectTableRow.js
Jonathanko52: For this one, I refered to the figma and checklist of action items.

#1646 

Screenshot 2024-09-04 at 10 39 47 AM

Link to prototype:

https://www.figma.com/proto/nD9QK56Mzq7xNSaSUoeGx0/TDM-Calculator?type=design&node-id=12173-2052&t=wQuxLI1JbOw0OzLn-1&scaling=scale-down&page-id=2499%3A11472&starting-point-node-id=11997%3A885

Issue #1578, opened 2024-01-14
Design Mockup: Redesign the My Projects Filtering UI


Biuwa:
### Overview
We need to redesign the filter user experience so that its self explanatory
### Action Items
- [x] ~Change the funnel icon (e.g to an arrow)~
- [x] ~Let the placement of the icon change with the user's interaction (e.g right facing to open, left facing to close~
- [ ] Reimagine how the filters could work

### Resources/Instructions
<img width="1284" alt="Screenshot 2024-01-14 at 10 13 15 AM" src="https://github.com/hackforla/tdm-calculator/assets/98936028/fe7b4125-f980-4ea4-b1a6-52f903d24976">

yiranshan:
1. Progress: communicated with teammates about how to approach this problem. Watched the weekly video recording and thought maybe the whole search panel should be moved to the side like the other pages. 
2. Blockers: Not sure what criteria are final for the project page. mock-up and test site should have different items. 
3. Availability: most nights from 9 pm to 11 pm. 
4. ETA: after the next weekly meeting, once all questions cleared and talked to teammates, we should be able to finish all 3 issues around search functions. 
![#1578 - ONGOING](https://github.com/hackforla/tdm-calculator/assets/79798670/049c9476-cf28-4141-acb7-8fd3cf46f071)
<img width="537" alt="Screen Shot 2024-01-27 at 2 07 11 PM" src="https://github.com/hackforla/tdm-calculator/assets/79798670/f2b0e571-6068-4420-b80f-a52f5cb11f5c">

yiranshan:
**Progress**
1. made a new filter button option based on the current design and setup. (see image 1 and 2)
2. designed a new version of searching and filtering inspired by LinkedIn and Google sheet (see image 3 and 4)
**Blockers**
1. still need to confirm what items would be included in the chart 
2. don't know if it's doable for dev team
**Availability**
6 hours min
**ETA**
hopefully the end of the week. 
![Filter Side Placement (#1579)](https://github.com/hackforla/tdm-calculator/assets/79798670/dc2eda0c-982f-4cfc-997f-bc5d254ad2da)
![Filter Side Placement (#1579) (1)](https://github.com/hackforla/tdm-calculator/assets/79798670/df2d7782-84dc-4c1c-83f0-8b089fd35ac8)

[...]

Issue #1817, opened 2024-09-05
Bug: Fix Universal Select Behavior


entrotech:
#### Describe the bug 

- Issue #1785 reported a bug where the appearance of drop-down lists varied among browsers and operating systems.
- PR #1802 was a first attempt to fix this by introducing a <UniversalSelect> component designed using the reactjs-select library, primarily to get control of the styling.

The reactjs-select library has no documentation I could find on how the value and onChange props can be used to make the drop-down act as a react controlled component.  This made the component behave differently than a standard react <select> component, and the confusion introduced bugs on the admin page, My Projects Filter Drawer and the Strategies page of the Calculation Wizard.

Issue #1816, opened 2024-09-04
Questions and Clarification of Design for My Project Page Filtering


entrotech:
### Dependency
- [x] When the dev team has new questions or clarifications from the design team
   - Add a question to the comment below 
   - Move the issue to the questions column
   - Add `ready for product` and `ready for design lead` label

### Overview
We need to review a list of potential improvements 

### Details
Yiran developed a re-design of the My Project Page filtering UI several months ago.  We are finally getting around to implementing the design, and have several questions and suggestions about the design that need to be resolved to implement the re-design.

### Action Items
- [x] Review the problems / Suggestions and decide which we will address and make issues 

### Problems / Suggestions based on developer review
- [x] #1829
<details><summary>Details</summary>

If you have all the popups closed, there is no indication of which, if any, filters are applied and no indication of any applied sorting.

<details><summary>Screenshots of TDM dev</summary>

I chose this filter

<img src="https://github.com/user-attachments/assets/e2a0cc48-1f14-443b-a5b4-92d22df85283" width=200px>

And then after I applied, it was not possible to see that I had used the filter

![image](https://github.com/user-attachments/assets/76b8636f-f4bd-49f7-a2da-6ceab4695d43)

</details> 

<details><summary>See Google Sheets filter indicator</summary>

![image](https://github.com/user-attachments/assets/ca25d7a2-e287-4f74-aaba-95b3845ece76)

</details> 

```
### Overview
As a user, I want to be able to see what filters and sorting I have choosen or is default so I understand my results better.

### Details
If you have all the popups closed, there is no indication of which, if any, filters are applied and no indication of any applied sorting.

### Action Items
- [ ]  Compare the current designs (in figma), to see if there was something that did not get developed and needs to be sent again to dev.
- [ ] If there is no design already, create one.

### Resources
- My Project's page https://tdm-dev.azurewebsites.net/projects

<details><summary>Screenshots of TDM dev</summary>

I chose this filter

<img src="https://github.com/user-attachments/assets/e2a0cc48-1f14-443b-a5b4-92d22df85283" width=200px>


And then after I applied, it was not possible to see that I had used the filter

![image](https://github.com/user-attachments/assets/76b8636f-f4bd-49f7-a2da-6ceab4695d43)

</details> 

<details><summary>See Google Sheets filter indicator</summary>

![image](https://github.com/user-attachments/assets/ca25d7a2-e287-4f74-aaba-95b3845ece76)

</details> 
```
</details> 

[...]

Issue #1785, opened 2024-08-08
Bug: Select Drop-Downs have different styling on different operating systems and browsers.


entrotech:
#### Describe the bug 

The select elements in the application have inconsistent appearance across browsers and do not adhere to the style guide.

#### Steps to reproduce the issue
- See Bonnie's comment on #1725 on august 1, 2024

#### What's the expected result?

The styling of select drop-down controls should be consistent.

#### What's the actual result?

- See Bonnie's comment on #1725 on august 1, 2024

#### Device configuration

Try different combinations of O/S:
Mac, Windows, Android, IOS

And different browsers: Chrome, Edge, Firefox, Safari (not Internet Explorer)

PR #1828, opened 2024-09-05
Implement filtering on Name, Address, Alt# and Created By column headings

### LINKED ISSUE #1647: Implement Filtering by Name, Address, Alt #, Building Permit, Created By from Column Headers on My Project Page
### Overview
- Implement the Name, Address, Alt #, Building Permit, Created By Filters on the My Projects Page per the Figma Prototype references in the Resources paragraph below to comply with the filter re-design of Issue #1578. 

### Action Items
- [x] This is partially implemented by Issue #1801. Please read that issue first for background information, adjustments John Darragh made to the design of all the filter pop-ups for consistency with our style guide, and information about what is done so far.
- [x] For this issue, you will need to complete the <TextPopup> component implementation.  You will probably need to modify one or more parent components to pass an additional prop that references the entire collection of projects, so you can extract the list of matches for the "auto-complete" component the design calls for.
- [x] Use the <DatePopup> component implementation in the same directory as an example.
- [x] remove the dependency from #1649 and move it to new issue approval column.
 
### Resources/Instructions

roslynwythe:
Fixes #1647

- The `ProjectTableColumnHeader` and `TextPopup` components are now called with a new prop `uniqueValues`, an array that includes sorted unique non-null values from all the currently displayed projects (all pages) for the specific project property.
- uniqueValues is used to populate the dropdown options in a UniversalSelect component, which wraps React-Select in order to provide consistent styling.
- The placeholder text "Search by keyword" was used based on the mockups in #1578.  A search icon is displayed to the left of the placeholder text, using same graphic used in the current search bar input.  Please let me know if a different icon or placeholder text should be used.  

- A new prop was required because the `TextPopup` component didn't have access to project properties which are required for generating match options.

<details>

image

</details>

<details>
  
image

</details>

### DIFF
--- a/client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
+++ b/client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
@@ -105,7 +105,6 @@ const TextPopup = ({
         disabled={false}
         onChange={e => {
           setNewSearchString(e.target.value);
-          console.log(e.target.value);
         }}
         value={newSearchString}
         defaultValue={newSearchString}
--- a/client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
+++ b/client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
@@ -5,6 +5,7 @@ import RadioButton from "../../UI/RadioButton";
 import "react-datepicker/dist/react-datepicker.css";
 import { MdClose } from "react-icons/md";
 import ReactSelect from "react-select";
+import SearchIcon from "../../../images/search.png";
 
 const TextPopup = ({
   selectOptions,
@@ -38,6 +39,17 @@ const TextPopup = ({
     setCheckedProjectIds([]);
     setSelectAllChecked(false);
   };
+
+  const placeholderComponent = (
+    <div>
+      <img
+        style={{ position: "absolute", left: "16 px", top: "14 px" }}
+        src={SearchIcon}
+        alt="Search Icon"
+      />
+      <div style={{ marginLeft: "30px" }}> Search by Keyword</div>
+    </div>
+  );
 
   return (
     <div style={{ display: "flex", flexDirection: "column" }}>
@@ -94,7 +106,7 @@ const TextPopup = ({
         }}
         defaultValue={newSearchString}
         styles={{ maxHeight: 200 }}
-        placeholder="Search by Keyword"
+        placeholder={placeholderComponent}
       ></ReactSelect>
       <hr style={{ width: "100%" }} />
       <div style={{ display: "flex" }}>
--- a/client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
+++ b/client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
@@ -4,8 +4,8 @@ import Button from "../../Button/Button";
 import RadioButton from "../../UI/RadioButton";
 import "react-datepicker/dist/react-datepicker.css";
 import { MdClose } from "react-icons/md";
-import ReactSelect from "react-select";
 import SearchIcon from "../../../images/search.png";
+import UniversalSelect from "../../UI/UniversalSelect.jsx";
 
 const TextPopup = ({
   selectOptions,
@@ -96,18 +96,22 @@ const TextPopup = ({
         }}
         value={newSearchString}
       /> */}
-      <ReactSelect
+      <UniversalSelect
         options={selectOptions.map(text => ({
           value: text,
           label: text
         }))}
+        name="inputName"
+        disabled={false}
         onChange={e => {
-          setNewSearchString(e.value);
+          setNewSearchString(e.target.value);
+          console.log(e.target.value);
         }}
+        value={newSearchString}
         defaultValue={newSearchString}
         styles={{ maxHeight: 200 }}
         placeholder={placeholderComponent}
-      ></ReactSelect>
+      ></UniversalSelect>
       <hr style={{ width: "100%" }} />
       <div style={{ display: "flex" }}>
         <Button onClick={setDefault} variant="text">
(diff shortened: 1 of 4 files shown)


client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
```diff
@@ -4,8 +4,11 @@ import Button from "../../Button/Button";
 import RadioButton from "../../UI/RadioButton";
 import "react-datepicker/dist/react-datepicker.css";
 import { MdClose } from "react-icons/md";
+import ReactSelect from "react-select";
```
entrotech, line 7: Long story: The styling of drop-downs was inconsistent across browsers and operating systems (Issue #1785). @bilalbg was not able to fix it with just CSS and resorted to introducing the react-select library to replace the standard react <select> component and creating a custom react component UniversalSelect which wrapped react-select and applied styles to get the appearance the designers wanted. Unfortunately, we dropped the ball on testing his PR, and though the list populated correctly, the selected value would not display the selected value, and this broke every page in the app with drop-downs.  I temporarily switched everything back to <select> to get through Tuesday's stakeholder meeting, but then looked into why the UniversalSelect wasn't working and created Issue #1817 and PR #1818 to fix it, and switched the existing usages back to <UniversalSelect>.  I revised the value and onChange props in UniversalSelect to work like they do in the <select> component. This may or may not be  the best option, but I was just trying to get it working for the usages that were in the app. At this time, we want all the drop-downs to use the UniversalSelect component, so we get the consistent styling. We can discuss if you want.

client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
```diff
@@ -67,15 +81,33 @@ const TextPopup = ({
...
         placeholder="Search by Partial Text"
         onChange={e => {
           setNewSearchString(e.target.value);
         }}
         value={newSearchString}
-      />
-
+      /> */}
+      <ReactSelect
+        options={selectOptions.map(text => ({
```
entrotech, line 100: I think the list would be more user-friendly if it was sorted alphabetically.

entrotech (changes requested):
Nice!  I made a few comments in the PR that I think should be addressed before merging.

Other thoughts about the design that I don't think should hold up merging your changes, but probably should be discussed with the designers:

1. The list often extends beyond the boundary of the drop-down window.  I had this problem with the date filtering popup as well.  Not sure whether we should do anything about that.
2. The prototype had checkboxes next to each item in the list - I guess they were thinking you might want to select multiple choices from the list, but  I'm not sure that's really needed.
3. It's also not clear to me whether you should be forced to pick one match from the list.  It might be useful to be able to just type a string, NOT select anything from the list, and apply the filter to get any projects where the column matches the substring.

client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
roslynwythe: Thank you @entrotech for fixing UniversalSelect.  The value and onChange props work fine, but since UniversalSelect does not accept a `placeholder` prop, I cannot place the search Icon next to the placeholder text.  Please advise.

client/src/components/Projects/ColumnHeaderPopups/TextPopup.js
roslynwythe: This is completed.

roslynwythe:
> Nice! I made a few comments in the PR that I think should be addressed before merging.
> 
> Other thoughts about the design that I don't think should hold up merging your changes, but probably should be discussed with the designers:
> 
> 1. The list often extends beyond the boundary of the drop-down window.  I had this problem with the date filtering popup as well.  Not sure whether we should do anything about that.
> 2. The prototype had checkboxes next to each item in the list - I guess they were thinking you might want to select multiple choices from the list, but  I'm not sure that's really needed.
> 3. It's also not clear to me whether you should be forced to pick one match from the list.  It might be useful to be able to just type a string, NOT select anything from the list, and apply the filter to get any projects where the column matches the substring.

- @entrotech I quoted your design related comments to #1816

roslynwythe:
> Nice! I made a few comments in the PR that I think should be addressed before merging.
> 
> Other thoughts about the design that I don't think should hold up merging your changes, but probably should be discussed with the designers:
> 
> 1. The list often extends beyond the boundary of the drop-down window.  I had this problem with the date filtering popup as well.  Not sure whether we should do anything about that.
> 2. The prototype had checkboxes next to each item in the list - I guess they were thinking you might want to select multiple choices from the list, but  I'm not sure that's really needed.
> 3. It's also not clear to me whether you should be forced to pick one match from the list.  It might be useful to be able to just type a string, NOT select anything from the list, and apply the filter to get any projects where the column matches the substring.

entrotech (commented):
After trying this out, I don't like how it works with a drop-down list box. 
1. You are forced to pick an item from the drop-down list, when I would prefer that you could just type a search string, and it would include any values that include the search string.
2. Once you select an item, and apply the filter, it still uses the the string selected as a substring in doing the matching, so does kind of match by a substring, but only lets you pick substrings that happen to be the complete string in one of the projects.
3. If you select an item to filter by, and then re-open the pop-up, you can't really unselect the pre-selected item in an intuitive way. You have to press Reset, then Apply to unlock the drop-down to show you other choices.

IMO, it was probably a mistake to use a listbox in the first place, and a better choice would be an auto-complete combo box, so the control allows any text, and the combo drop-down is just a convenience, in case they happen to be looking for an exact match.

We can do any or all of the following:
Merge this as-is and then we can demonstrate the problems to the PMs and design team. 
We could go off-script and come up with something we like and propose that to the team as an alternative.
We could just kick this back to the design team with comments about what we don't like and see if they can come up with something better.

What do you think?

roslynwythe:
@entrotech Regarding 1 and 2, do you think we could use [async](https://react-select.com/home#async) to enable the user to type a search string and then have matching options load, then the user would have the option of selecting from the list?

I agree that clearing the filter is a bit awkward and it might be worthwhile to consider a design change.

If you wish to merge the PR so that it is more readily available for demonstration, that is fine, but I'm also happy to continue working on it until it meets the customer's needs.

PR #1831, opened 2024-09-07
Filters implemented with new state, different selectors

### LINKED ISSUE #1646: Implement the Visibility and Status Column Filters on the My Project Page

Jonathanko52:
Fixes #1646

-Kept state in the visibility and status popups.
-Used our own Radio button component and a custom checkbox, instead of Universal Selector.
-Removed "Deleted" keyword from the modified date column
-Added "Deleted" keyword to status column and background color for deleted projects. I'm not particularly attached to this change, but it's an action item in the description for this issue, so I'm including that for now.

In my previous pull request, I'd created too many state/prop items and drilled them through multiple components, creating a lot of entanglement and possible points of failure. This version keeps changes limited to the popup components. It also uses radio buttons and checkboxes instead of dropdowns, keeping it closer to the prototype.

Screenshot 2024-09-07 at 12 03 06 AM
Screenshot 2024-09-07 at 12 03 10 AM
Screenshot 2024-09-07 at 12 03 03 AM

### DIFF
--- a/client/src/components/Projects/ColumnHeaderPopups/StatusPopup.js
+++ b/client/src/components/Projects/ColumnHeaderPopups/StatusPopup.js
@@ -20,12 +20,16 @@ const StatusPopup = ({
     header.id !== orderBy ? null : order
   );
 
+  const [typeSetting, setTypeSetting] = useState(criteria.type);
+  const [showDeleted, setShowDeleted] = useState(criteria.status === "all");
+
   // TODO More state variables for status filtering go here
 
   const setDefault = () => {
     setCriteria({
       ...criteria,
-      [header.id]: ""
+      status: showDeleted ? "all" : "active",
+      type: "all"
     });
     setCheckedProjectIds([]);
     setSelectAllChecked(false);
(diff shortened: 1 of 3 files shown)


entrotech (approved):
This is great - thank you!  I had forgotten about the part about moving the deleted text to the status column and changing the background color of the deleted project rows.

PR #1934, opened 2024-11-04
update classes to prevent cutoff of left side of project table

### LINKED ISSUE #1927: Bug: When browser window is not extra wide, cannot see the left-most columns on My Projects Page
#### Describe the bug 

When the browser window is not extra wide, there is a horizontal scroll bar, but the user cannot scroll far enough to the left to reveal the left-most columns.

#### Steps to reproduce the issue

(if applicable, please specify platform (iOS, Android, Windows, Mac version) and brower)

REPLACE THE FOLLOWNG TEXT:
- 1. Go to the My Projects Page
- 2. Narrow your browser window
- 3. There is a horizontal scroll bar, which is scrollled to the left, but the left-most columns are off-screen to the left.

#### What's the expected result?
You should be able to scroll to see any of the columns.

#### What's the actual result?
Cannot scroll horizontally to see left-most columns.

image

#### Device configuration

At least on this hardware/software:
- Device: laptop
- OS version: Windows 11
- Browser: Chrome and Firefox

#### Closing
- [ ] Review with product and dev. Update if needed based on feedback
- [ ] Once finalized, add before and after images to the [staging deck on this slide](https://docs.google.com/presentation/d/1crZ3IxqA4hAu3qzD7ns93Ieuqjwh6wyEtuX_46cP-fg/edit#slide=id.g3163b6ed272_0_0) 

roslynwythe:
Fixes #1927

- use JavaScript to modify style of the `body` element in `<ProjectTable>`, to prevent page-wide scrolling
- enable scrolling on `div.tableContainer`
- make table headings "sticky" so they are always visible regardless of vertical scroll position
- changes to the zIndex (stacking order) of header and popover so that popovers appear "on top" of the table heading.   
- change minHeight on contentContainer because the previous value did not adjust with user's selection of font size
- At screen widths less than 800px, the toolbar, input field, and button directly above the table become cut off.  Regardless of how scrolling is implemented, the design and implementation of those elements should be revisited to be responsive at smaller screen sizes.  

- When the browser window was not extra wide, there was a horizontal scroll bar, but the user could not scroll far enough to the left to reveal the left-most columns.  
- the requirement to make table headings "sticky" was not in the issue but was requested in [comment]( https://github.com/hackforla/tdm-calculator/pull/1934#issuecomment-2458583512)

<details>

image

</details>

<details>
  
image

</details>

### DIFF
--- a/client/src/components/Layout/ContentContainer.jsx
+++ b/client/src/components/Layout/ContentContainer.jsx
@@ -7,7 +7,7 @@ const useStyles = createUseStyles({
     flex: "1",
     display: "flex",
     flexDirection: "row",
-    minHeight: "calc(100vh - 103px)"
+    minHeight: "calc(100vh - 67px - 5em)"
   },
   content: {
     boxSizing: "border-box",
--- a/client/src/components/Layout/ContentContainer.jsx
+++ b/client/src/components/Layout/ContentContainer.jsx
@@ -7,7 +7,7 @@ const useStyles = createUseStyles({
     flex: "1",
     display: "flex",
     flexDirection: "row",
-    minHeight: "calc(100vh - 67px - 5em)"
+    minHeight: "calc(100vh - 71px - 5em)"
   },
   content: {
     boxSizing: "border-box",
--- a/client/src/components/Layout/ContentContainer.jsx
+++ b/client/src/components/Layout/ContentContainer.jsx
@@ -7,7 +7,7 @@ const useStyles = createUseStyles({
     flex: "1",
     display: "flex",
     flexDirection: "row",
-    minHeight: "calc(100vh - 103px - 48px)"
+    minHeight: "calc(100vh - 103px)"
   },
   content: {
     boxSizing: "border-box",
--- a/client/src/components/Layout/ContentContainerNoSidebar.jsx
+++ b/client/src/components/Layout/ContentContainerNoSidebar.jsx
@@ -9,7 +9,8 @@ const useStyles = createUseStyles({
     justifyContent: "flex-start",
     alignItems: "center",
     minHeight: "calc(100vh - 103px - 48px)",
-    margin: "auto"
+    margin: "auto",
+    width: "100%"
   }
 });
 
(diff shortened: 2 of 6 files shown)


entrotech:
This fix introduces another problem when you use the horizontal scroll bar and scroll to the right, the page is wider than the headers at the top of the page:

image

I think what we probably need to do is enclose the gird in a div that allows horizontal scrolling, instead of having the scroll bar apply to the whole page. Others mentioned last week that the page heading and all the stuff just above the  grid should stay on-screen regardless of scroll position.

entrotech (changes requested):
This introduces another problem - see my comment on the PR with a screen shot.

roslynwythe:
Update:
By applying the following changes, I was able to achieve the desired behavior:
- remove `overflow-y` from body 
- add `overflow: auto` and `max-width: 100vw` to class tableContainer 

However, there is a comment relating to the style `overflow: auto` on tableContainer.  The comment states:
```
changed to allow Universal Select to show above the page container when expanded
```
which came from 
https://github.com/hackforla/tdm-calculator/commit/0c76abe85e4107346a89d01f52e111cebbe9e94a#diff-33d73623983cfa22c6e9b767d1648a9f9fe52003012845d0a5219c1afdfbed68R121

and I am not sure how to test this.

roslynwythe:
@entrotech I wm working on two options:
1. On the Projects Page, implement scroll bars at the table Container level instead of at the page level.  This can be done by using JavaScript in the <ProjectPage> component, to override the style on the body tag, inhibiting page level scrolling.  However, at screen sizes < 850 px, some problems develop with some elements on the toolbar above the table.  Assuming that we are supporting tablet user, these problems probably should be addressed, regardless of the outcome of this issue, otherwise users will either have tools cut off or have unnecessary and awkward scrolling.    

2. Continue to allow the default page-level scrollbars, but use advanced techniques to match the width of the global la header and the tdm header and footer to the width of the table.  The hook `useMeasure()` could be used to dynamically read the width of the table and the set the width of the headers and footers to match, in order to avoid the appearance shown in your screenshot above.

roslynwythe:
@entrotech Thank you for merging this PR.  You may have noticed that at smaller screen sizes, some features (of the toolbar above the table) are cut off, and I believe that some design and implementation changes should be considered to make those elements more responsive.   

I have an alternative solution in which scrolling is done at the page level, and the problem depicted in https://github.com/hackforla/tdm-calculator/pull/1934#issuecomment-2458583512 is resolved by the use of CSS and JavaScript to dynamically adjust the width of the various elements to align.

PR #2141, opened 2025-02-28
fix: warning tooltip no longer fixed | prop cleanup

### LINKED ISSUE #2042: Bug: Warning tooltip has fixed position
#### Description

The warning tooltip has its position as fixed, causing it not to stay with the Login button when the user scrolls down the page.

#### Steps to Reproduce
1. Navigate to the TDM Login page.
2. Click the Continue Without Saving button.
3. Notice the warning tooltip that appears below the Login button in the top-right corner.
4. Scroll down the page.

#### Expected Result
The warning tooltip should scroll with the Login button.

#### Actual Result
The warning tooltip remains fixed in the same position on the viewport.
<details><summary>Warning tooltip fixed position</summary>

Screenshot 2025-01-15 170843
</details>

#### Device configuration
- Device: Desktop
- OS version: Windows 11
- Browser: Firefox
- Browser version: 134.0.1

#### Action Items
- [ ] Fix bug.
- [ ] Review with product and dev. Update if needed based on feedback.
- [ ] Once finalized, add a slide to the [the staging deck](https://docs.google.com/presentation/d/1crZ3IxqA4hAu3qzD7ns93Ieuqjwh6wyEtuX_46cP-fg).

Tony-Villa:
Fixes #2042 

- Made it so warning tooltip on landing page is no longer fixed on scroll
- Removed a prop `lockScroll` from `<Popup>` because that prop only works with modals not tooltips

- Tooltip on landing page was fixed on scroll, it would move with the scroll instead of staying anchored to it's trigger. In this case the `login` link
- The prop cleanup is not user facing, just some cleanup to try to avoid any tech debt

- N/A

<details>

https://github.com/user-attachments/assets/45da98cc-ddcf-41b1-9d2e-4bcf99a91037

</details>

<details>
  

https://github.com/user-attachments/assets/bbb10abc-6e2b-44c7-b153-1c0345aed71d

</details>

### For Devs
This tooltip is using the library [reactjs-popup](https://github.com/yjose/reactjs-popup), according to this [issue](https://github.com/yjose/reactjs-popup/issues/208#issuecomment-692328775) with the latest major release this became an internal bug in the library. 

What I did to remedy this was add an event listener to the body element and used an effect to move the tooltip on scroll. This is kind of a hacky way to fix this, which is why it's a tiny bit floaty, but it fixes the bug. We can revisit this once it has been fixed in the library. I made sure to remove the listener when the effect is cleaned up.

### DIFF
--- a/client/src/components/Layout/NavBarLogin.jsx
+++ b/client/src/components/Layout/NavBarLogin.jsx
@@ -71,6 +71,26 @@ const NavBarLogin = ({ classes, handleHamburgerMenuClick }) => {
     </li>
   );
 
+  useEffect(() => {
+    const handleScroll = () => {
+      const tooltip = document.querySelector(".popup-content");
+      const loginButton = document.getElementById("cy-login-menu-item");
+      if (tooltip && loginButton) {
+        const rect = loginButton.getBoundingClientRect();
+        tooltip.style.top = `${Math.floor(
+          rect.bottom + window.scrollY + 10
+        )}px`;
+      }
+    };
+
+    const scrollableElement = document.querySelector("#body");
+    scrollableElement.addEventListener("scroll", handleScroll);
+
+    return () => {
+      window.removeEventListener("scroll", handleScroll);
+    };
+  }, []);
+
   return !account || !account.email ? (
     !isCalculation ? (
       <li className={clsx(classes.userLogin, classes.linkBlock)}>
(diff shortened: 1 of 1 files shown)


entrotech (approved):
This is great. Cool that it was a bug in the library that they plan to fix!

PR #2312, opened 2025-04-02
add projectHidden table and update associated stored procedures

### LINKED ISSUE #2163: Hiding Projects Should be User-Specific
### Overview
When you hide a project, you should only be hiding it from yourself, and anybody else who has access to the project can hide it from themself, but not from the author or anyone else. This allows each user with access to a project to decide for themself if a project should be hidden.

### Explanation
Currently, only a project author can Hide a project, and it will become hidden for everyone. Other users can access a project if:

- they are an admin, or
- the project is a snapshot and the author shared it with them.

It is desirable for you to hide projects belonging to others, and conversely when you hide one of your own projects for other to still be able to see it. Some relevant use cases:

- An administrator decides that they no longer need to see obsolete projects.
- Projects that are shared with a non-admin user have been copied to drafts that now belong to the user, so they can hide obsolete shared snapshots.

### Action Items
- [x] Create a database table named ProjectHidden with columns projectId int FK to the project table, dateHidden datetime2(7) not null, loginId int FK to Login table.
- [x] Modify ProjectHide stored procedure to insert or delete records into the ProjectHide table instead of updating the dateHidden field in the Project table.
- [x] Modify the ProjectSelect* stored procedures to return dateHidden by doing an outer join to the ProjectHidden table for the user's loginId instead of getting the dateHidden from the Project table itself.
- [x] I don't believe any front-end changes are required, unless we decide to present the user with an explanation of how hiding only hides a project from them, and not from other users who might have access to the project.
- [ ] Review with product and dev. Update if needed based on feedback
- [ ] Once finalized, add before and after images to the [staging deck on this slide]() 
   - [ ] if no slide has been made, make one in [the staging deck](https://docs.google.com/presentation/d/1crZ3IxqA4hAu3qzD7ns93Ieuqjwh6wyEtuX_46cP-fg)

### Resources/Instructions
- [2025-04-15 Release Deck, slide 2163](https://docs.google.com/presentation/d/1_Xszms-WP5drcEXG3E_d8EiJ7qqtgqehZ1-Hw6p1xCw/edit#slide=id.g34ad0e17a79_0_416)

ale210:
- Closes #2163

- added `ProjectHide` table to track hidden for individual users instead of global
- updated stored procedures to use this new table

- hiding a project should be user specific, not global. one project hidden by one user should still be visible to other users


server/db/migration/V20250401.1526__add-project_hidden.sql
```diff
@@ -0,0 +1,264 @@
...
+			, p.dateTrashed
+			, p.dateSnapshotted
+			, p.dateSubmitted
+         , p.droId  -- New column
+         , p.adminNotes  -- New column
+         , p.dateModifiedAdmin  -- New column
+      FROM Project p
+      JOIN Login author ON p.loginId = author.id
+	   LEFT JOIN ProjectHidden ph on p.id = ph.projectId
+      WHERE author.id = ISNULL(307, author.id) OR EXISTS
```
fyliu, line 119: I'm not sure what this is doing

> author.id = ISNULL(307, author.id)

Looks like it's the same as 

> author.id = 307

It used to be like this

> author.id = ISNULL(@loginId, author.id)

Maybe 307 is a specific user you used to figure out the sql for this?

fyliu (changes requested):
I'm very rusty and wasn't that good with sql to begin with. I found one thing and thought I should post it while I keep checking/comparing the other procedure changes.

server/db/migration/V20250401.1526__add-project_hidden.sql
ale210: @fyliu ah this is actually from a [previous commit](https://github.com/hackforla/tdm-calculator/commit/2b2ce47269a1e05966360ce1cfc0472a4e7ee43d)

it looks like it substitutes 307 if a project has no author?

entrotech: No, the 307 should be @loginId. I fixed it in a subsequent DB migration that probably had not been merged when Alex pushed his PR. I'm in the process of sorting it out, cause we need the adjustments Alex made plus the changes I was making that allow users to see snapshots shared with them.  I'm working on a combined fix, but just haven't been able to  get around to finishing it just yet. Should be fixe this afternoon.  Thanks for letting me know, since I was just going to mod Alex's migration script, but I should just make a newer script to fix it up.

entrotech: I also need to do some front-end changes that will permit users to hide projects that do not belong to them for this feature to be accessible.

Issue #2404, opened 2025-04-18
Dev: Spelling: Exclude list: Initial Export all text from the database (tooltips, etc.)


lola3736:
### Overview
We need to find a way to process strings saved to the database with the spell checker.

### Details 
- Many of the strings displayed to users on the application are stored in the database, so we want to find a way to run a spell check on these strings.
- The fields we want to check are populated by database migration scripts, so they should be the same on the dev environment as in production, so it may be sufficient to run this process only against the development database for now.  At some point in the future, production admin users are likely to edit the production FAQ data, so we will need a way to also run spell checking on the FAQs against the production database eventually.
- One approach is to run a web api request to dump all the strings in some fashion into a file (probably JSON), then run the existing spellcheck script and include the file in the files that get parsed by the spell checker.
- This should be an ongoing process.  If there is a way to automate it, please detail that in the issue.  
- If it will be manual, please detail the steps so that a more junior person can repeat the process later.
- Once the content has been exported, ~we just need a list of the unique words.~ we need all the places/context the words appear so that we can change them later.
- ~There is another issue to add it to the database after it's exported #2405~ These have been comboed in to this issue
- We need to upload the text content from the CSV or JSON from the database export to the repo
- When we upload the file to the repo via a PR the linter will run and let us know what words it identifies as "misspelled"
- please download the list it prints to the console so that we can review to find actual misspellings and identify what additional words to add to the exclude part of the linter
- use the template issue to report words that need to be added to Exclude list

### Action Items
- [x] Assign yourself to the development issue.
- [x] Move the development issue from the Prioritized Backlog to In Progress.
- [x] Decide on  an approach.
- [x] Implement a script that  runs the checker.
- [ ] Modify the [spell checking wiki page](https://github.com/hackforla/tdm-calculator/wiki/Spell-Checking) to include any additional steps required to include the database in the spell checking procedure.
- [ ] Create a PR, if applicable.
- [ ] Once the PR has been approved, use it to close this issue. If there is no PR associated with this issue, close this issue.
- [ ] Release the dependency on #2405

#### Slides
- [Staging Deck](https://docs.google.com/presentation/d/1crZ3IxqA4hAu3qzD7ns93Ieuqjwh6wyEtuX_46cP-fg)
[...]

PR #2466, opened 2025-04-25
2404 db spellchecking

jackberrypassionfruit:
- Fixes #replace_this_text_with_the_issue_number

Added a script to pull tables from the database and run a spellchecker script on them to check for mistakes. This check has been added as an npm script within `./server/node_modules`.

- The database data is not easily visible to most developers, so automating spellchecking for it is helpful to be able to notify database administrators and submit new issues accordingly.

_Note, I ran into an issue with the linter regarding CL/LF characters while committing this PR, potentially because I am developing via WSL2 on a Windows machine. This has led to ~30 files being affected, which was not my intent, so hopefully this can be resolved before any changes are potentially made to the "develop" branch._


.gitignore
ryanfchase, line 1: I'm not sure why this got marked as changed. Could you try checking out this file so it matches develop, and so it won't show up as modified by you?

.vscode/settings.json
```diff
@@ -1,6 +1,9 @@
...
-  "editor.formatOnPaste": true,
-  "files.eol": "\n"
-}
+{
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true,
+  "editor.formatOnPaste": true,
+  "files.eol": "\n",
+  "[javascriptreact]": {
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
ryanfchase, line 7: Is this required for your feature to work? Otherwise, I think we should leave it out. See my comment about formatting if this was added as a response to the pre-commit hook problem.

ryanfchase (changes requested):
I'm seeing a number of files with changes to indentation (primarily with nested ternary operators). I assume you did this because of the pre-commit hook that complains when you try to commit? I was able to resolve this by following instructions on the wiki [Contributing Code](https://github.com/hackforla/tdm-calculator/wiki/Contributing-Code), namely installing node packages in the client folder with `--legacy-peer-deps` options. Can you try undoing these indentation changes and see if those instructions fix this issue?

ryanfchase:
Also, regarding your file endings comment. I am only really seeing that for the new files that you've committed. It doesn't seem to have affected files that you simply modified. It seems that among the other devs here, simply adding an extra newline to the end of your files will be good enough. Feel free to take that advice with a grain of salt.
- your comment: https://github.com/hackforla/tdm-calculator/issues/2404#issuecomment-2829192077

entrotech (commented):
Can you see if you can fix the line-ending problem?  I run on Windows using Git Bash for my terminal.  [This wiki page](https://github.com/hackforla/tdm-calculator/wiki/Git-for-Windows-Installation) shows that there is a critical setting in setting up Git Bash that sets it up to handle line endings correctly when the repo is used cross-platform.  WSL may have a similar configuration setting.

.gitignore
entrotech: Looks like Jack added line 9, so git would ignore the downloaded database data.  The rest of the file is likely just a difference in line endings.

.vscode/settings.json
entrotech: This is the likely cause of all the erroneous changes to the files where the indenting of ternary statements got modified.  Prettier changes their mind about three times over the last few years about how multiline ternaries should be formatted, but this change looks like it would substitute a different formatter that is unlikely to use the same formatting rules as prettier.

.gitignore
entrotech: I also noticed that this new line in the .gitignore didn't seem to prevent the csv files from being included in the PR.  This might be a "chicken and egg" situation where the .gitignore needs to be added to the repo before it has any effect in excluding the files that don't need to be in the repo?

server/db/spelling_test/spelling_test.js
```diff
@@ -0,0 +1,79 @@
...
+        ${exclude_where}
+        
+    ) AS x 
+    `);
+
+  schema_result.recordset.forEach(async row => {
+    const { table_name, column_names } = row;
+
+    const table_result = await request.query(
+      `SELECT ${column_names} FROM ${table_name}`
```
entrotech, line 57: Would it be helpful to include the "Id" column in this query as well, so we know which row in the table has the spelling error?

package.json
entrotech, line 1: Seems like a lot of NPM packages to add here. Are they all actually used?

entrotech (changes requested):
This is really cool!  I'd like to see it cleaned up a bit, as far as changes that seem to just seem to be side-effects of your dev environment.

.gitignore
jackberrypassionfruit: That's a good catch, John. I'll remove those .csv's from the commit that I add to this PR.

.vscode/settings.json
jackberrypassionfruit: At some point I had downloaded a module to be able to run `npx eslint --fix`, and I may have not covered all my tracks. I will remove that setting from this file.

package.json
jackberrypassionfruit: "mssql-to-csv" is a module that turned out to be unsupported, and "prettier-eslint" was the tool I installed to fix my linter errors. I can remove them both from this PR.

"@cspell/cspell-json-reporter" is a tool I used to export the CSpell warnings to a CSV, it turned out to be a pain to implement, but I left it in case that functionality was requested after this PR.

server/db/spelling_test/spelling_test.js
jackberrypassionfruit: This is a good idea, thanks for the reminder. I will update the query so it pulls whatever column has the attribute "is_identity=1", otherwise it won't show look for that column.

entrotech (approved):
Very cool!   I had to keep the folder db_table_csv for the server script to be able to write to the files, but then everything seemed to work well. I'll create a follow-up issue to do the db migration to fix the few spelling errors - even though none of them actually show up in the UI.

jackberrypassionfruit:
Ah, that's true. A quick check for `!fs.existsSync(dir)` would indeed be a bit more robust.

Glad to see this issue resolved. I hope it makes a good template for further projects. Thank you John.

PR #2723, opened 2025-10-18
Tv/a11y projects page

Tony-Villa:
- Fixes N/A (no-ticket)

- This handles all the accessibility errors from the WAVE chrome extension.

- Be a11y compliant

<details>

Screenshot 2025-10-18 at 3 01 20 PM

</details>

<details>
  
Screenshot 2025-10-18 at 2 55 56 PM

</details>

### Dev Notes:
I added a global `sr-only` (screen reader only) which will be useful for anyone else working on fixing these issues.

### DIFF
--- a/client/src/components/Projects/ProjectsPage.jsx
+++ b/client/src/components/Projects/ProjectsPage.jsx
@@ -1156,7 +1156,7 @@ const ProjectsPage = ({ contentContainerRef }) => {
                     placeholder="Search by Name; Address; Description; Alt#"
                     value={filterCriteria.filterText}
                     onChange={e => handleFilterTextChange(e.target.value)}
-                    aria-roledescription="Search for project"
+                    aria-label="Search for project"
                   />
                   <MdOutlineSearch className={classes.searchIcon} />
                 </div>


client/src/components/Projects/ProjectsPage.jsx
```diff
@@ -1127,6 +1145,9 @@ const ProjectsPage = ({ contentContainerRef }) => {
                 }}
               >
                 <div className={classes.searchBarWrapper}>
+                  <label htmlFor="filterText" className="sr-only">
+                    Search the site
```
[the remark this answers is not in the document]
Tony-Villa: yep, you're right it does make more sense to say that. Will change. Good catch!

client/src/components/Projects/ColumnHeaderPopups/ProjectTableColumnHeader.jsx
```diff
@@ -359,8 +359,10 @@ const ProjectTableColumnHeader = ({
             orderBy={orderBy}
           ></ColumnHeader>
         </Popover>
-      ) : (
+      ) : header.label ? (
```
ryanfchase: needs to be `header.label` is a string? then use it in the span.

Otherwise, let's add another prop to `headerData` called `ariaLabel`:

```
   {
      id: "checkAllProjects",
      ariaLabel: 'select all projects'
      label: (
        <input
          style={{
            height: "15px"
          }}
          type="checkbox"
          checked={selectAllChecked}
          onChange={handleHeaderCheckbox}
        />
      ),
      colWidth: "3rem"
    },
// ...
    {
      id: "contextMenu",
      ariaLabel: 'project options',
      label: "",
      colWidth: "3rem"
    }
```

client/src/components/Projects/MultiProjectToolbarMenu.jsx
```diff
@@ -216,15 +219,18 @@ const MultiProjectToolbarMenu = ({
             }`}
             disabled={isHideBtnDisabled}
             onClick={handleHideBoxes}
+            aria-label="Toggle Visibility"
```
ryanfchase: aria-label should change based on `dateHidden`: "Make Visible" vs "Make Hidden"

client/src/components/Projects/MultiProjectToolbarMenu.jsx
```diff
@@ -170,8 +171,10 @@ const MultiProjectToolbarMenu = ({
             }`}
             onClick={handlePrintPdf}
             disabled={checkedProjectIds.length !== 1}
+            aria-label="Print PDF"
           >
             <MdPrint
+              aria-hidden="true"
               className={
```
ryanfchase: add `aria-disabled` so that screen reader will read out if a button is disabled. set it to `checkedProjectIds.length !== 1`

client/src/components/UI/TableData.jsx
```diff
@@ -98,13 +98,21 @@ const TdExpandable = ({ children }) => {
         </span>
         {isTruncated && (
           <button
+            aria-expanded={isExpanded}
```
ryanfchase: can you explain which error this fixes?

ryanfchase (changes requested):
Requesting explanations so that we can map fixes to WAVE errors (we will want to make Wiki page entries so that other devs can make use of your fixes). 

There was one change needed for the `headerData` array, let us know if there are any questions.

client/src/components/Projects/ColumnHeaderPopups/ProjectTableColumnHeader.jsx
Tony-Villa: Thanks for the tip, went ahead and did that and added a new fallback in the case of new columns being added

PR #2786, opened 2025-11-14
feat: add custom hook to handle reactjs-popup attribute stripping issue

Tony-Villa:
- Fixes N/A

- Added a custom hook to abstract the reactjs-popup library attribute stripping issue

- This makes it more reusable and less rewriting of similar code throughout the codebase 

I've added documentation on how to use this hook for people working on the accessibility issues. Linked below:

https://github.com/hackforla/tdm-calculator/wiki/Accessibility-Audits#custom-hooks-documentation


client/src/hooks/useRemovePopupAriaIssue.js
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from "react";
+
+export const useRemovePopupAriaIssue = ({ elementId, deps, setControls }) => {
```
geolunalg: Love this! ❤️ 

Can we rename the file/hook to `useAccessibilityAriaPopup`? Also, it would be great if you could provide top-level function comments so the purpose is easy to understand. See my example below:

```
/**
 * Custom React hook that addresses accessibility issues by providing
 * descriptive text to an ARIA popup so screen readers can read it to the user.
 *
 * @param {string} elementId - The ID of the element the popup is applied to.
 * @param {number} deps - Dependencies that trigger updates for the hook.
 * @param {boolean} [optionalValue] - Optional value used to control behavior.
 * @returns {void} This hook does not return any value directly.
 */
export const useAccessibilityAriaPopup = ({ elementId, deps, setControls }) => {
    ...
}
```

PR #3453, opened 2026-08-19
Fix project snapshot duplication with corrective code refactoring

### LINKED ISSUE #3353: Bug: 'CREATE A COPY' Button on the Duplicate Project Modal is Malfunctioning
#### Describe the bug
The 'CREATE A COPY' button on the Duplicate Project modal is unresponsive when trying to duplicate a snapshot owned by the user.

#### Steps to Reproduce
1. Log in as User
2. Go to the My Projects page
3. Filter Status column by Snapshot
4. Select a snapshot owned by User
5. Go to the kebab menu on the right of the snapshot
6. Select 'Duplicate'
7. On the modal that pops up, click 'CREATE A COPY'
8. Observe bug

#### Actual Result
The 'CREATE A COPY' is unresponsive, only 'CANCEL' works

#### Expected Result
Both 'CREATE A COPY' and 'CANCEL' should respond to the user clicking on them

#### Device Configuration
- Device: n/a
- OS version: n/a
- Browser: n/a
- Browser version: n/a

#### Action Items
- [ ] Fix bug.
- [ ] Review with development lead.
- [ ] Review with product, development, and research teams. Update if needed based on feedback.
- [ ] Once finalized, add a slide to the staging deck (see resources) 

#### Staging and Release decks
- [Staging Deck](https://docs.google.com/presentation/d/1crZ3IxqA4hAu3qzD7ns93Ieuqjwh6wyEtuX_46cP-fg/)
   - ~[Staging slide, 3353](https://docs.google.com/presentation/d/1crZ3IxqA4hAu3qzD7ns93Ieuqjwh6wyEtuX_46cP-fg/edit?slide=id.g3f2afe8206d_0_0#slide=id.g3f2afe8206d_0_0)~
- Release Deck
   - [2026-09-01 Release deck, slide 3353](https://docs.google.com/presentation/d/1uQpewpS226aQFV7ekeZkRzRAcXgvrjvMAE5iRQJt9Zw/edit?slide=id.g3fadd5c5e52_0_0#slide=id.g3fadd5c5e52_0_0)

ijoel:
- Fixes #3353

The 'CREATE A COPY' Button on the Duplicate Project Modal triggers a handling function that implements the snapshot duplication for the "Projects Page"

[Issue 3353](https://github.com/hackforla/tdm-calculator/issues/3353) indicates that an unresponsive failure or in-action occurs when clicking the "CREATE A COPY" button.

After review, a set of fixes were identified, fixed, and the code was rearranged and refactored to reduce intermediate variables and made more tidy. Duplicating a snapshot on the projects page is fixed by these chages.

The handling function had the following errors:

1. Passing a project object as an argument to `projectService.getById` that expected a number value

2. Target point calculation, `admin` only, was missing an authorization guard

3. The project object was being mistakenly derived (shallow copied) from an `axios` response instead of the desired nested `data` property.

### DIFF
--- a/client/src/components/Projects/ProjectsPage.jsx
+++ b/client/src/components/Projects/ProjectsPage.jsx
@@ -418,23 +418,22 @@ const ProjectsPage = ({ contentContainerRef }) => {
   });
 
   const handleCopyModalClose = async (action, newProjectName) => {
-    let newSelectedProject = { ...selectedProject };
     if (action === "ok") {
-      const projectFormInputsAsJson = JSON.parse(selectedProject.formInputs);
-      projectFormInputsAsJson.PROJECT_NAME = newProjectName;
-      if (!selectedProject.targetPoints) {
-        await projectResultService.populateTargetPoints(selectedProject);
-        newSelectedProject = await projectService.getById(selectedProject.id);
+      if (isAdmin && !selectedProject.targetPoints) {
+        await projectResultService.populateTargetPoints(selectedProject.id);
       }
-      let newProject = {
-        ...newSelectedProject,
+      const projectFormInputsAsJson = {
+        ...JSON.parse(selectedProject.formInputs),
+        PROJECT_NAME: newProjectName
+      };
+      const { data } = await projectService.getById(selectedProject.id);
+      const newProject = {
+        ...data,
         loginId: loginId,
         name: newProjectName,
+        description: data.description ?? "",
         formInputs: JSON.stringify(projectFormInputsAsJson)
       };
-      if (!newProject.description) {
-        newProject.description = "";
-      }
       try {
         await projectService.post(newProject);
         await updateProjects();


client/src/components/Projects/ProjectsPage.jsx
```diff
@@ -418,23 +418,22 @@ const ProjectsPage = ({ contentContainerRef }) => {
...
 
   const handleCopyModalClose = async (action, newProjectName) => {
-    let newSelectedProject = { ...selectedProject };
     if (action === "ok") {
-      const projectFormInputsAsJson = JSON.parse(selectedProject.formInputs);
-      projectFormInputsAsJson.PROJECT_NAME = newProjectName;
-      if (!selectedProject.targetPoints) {
-        await projectResultService.populateTargetPoints(selectedProject);
-        newSelectedProject = await projectService.getById(selectedProject.id);
+      if (isAdmin && !selectedProject.targetPoints) {
```
entrotech, line 422: This conditional logic handles a situation caused by the addition of a few columns in the project table that store previously computed results `targetPoints`, `earnedPoints` and `projectLevel`. Since these are only computed by the client, it was not possible to run a script to update these columns for projects that existed prior to the new columns to populate them with values with a simple script, so this conditional logic is a hack to get the client to compute and store these values in the db if the column values are null.

Since you created this PR from your personal forked repo instead of a feature branch within the tdm-calculator repo, I'm not able to just modify a feature branch to update the PR, but will need to create a separate new PR to make two small adjustments to the conditions for running the populateTargetPoints method:
- The method will also need to be called by the author of the project, even if they are not an admin, and
- A project can now be saved with a targetPoints value of 0, which is also a falsey value, but in this case, the method does not need to be called.

entrotech (approved):
I'm approving,  but will need to create another follow-up PR to make a small adjustment as noted below.

